PR #6, opened 2023-03-17
Do not throw exception when git installation is missing

bernhardschaefer:
# Summary

When git is not installed the import of foundry_dev_tools fails with the stack trace below. As the stacktrace shows, the  issue is that foundry_dev_tools uses the git cli to try to find a .foundry_dev_tools folder in the project root directory.
This bugfix follows the approach in transforms.api._dataset._get_branch and also catches the FileNotFoundError.

```shell
  File "/opt/conda/lib/python3.10/site-packages/foundry_dev_tools/__init__.py", line 9, in <module>
    ) = foundry_dev_tools.config.initial_config()
  File "/opt/conda/lib/python3.10/site-packages/foundry_dev_tools/config.py", line 152, in initial_config
    project_config_file = _find_project_config_file(
  File "/opt/conda/lib/python3.10/site-packages/foundry_dev_tools/config.py", line 349, in _find_project_config_file
    git_directory = _traverse_to_git_project_top_level_dir(project_directory)
  File "/opt/conda/lib/python3.10/site-packages/foundry_dev_tools/config.py", line 326, in _traverse_to_git_project_top_level_dir
    return execute_as_subprocess(["git", "rev-parse", "--show-toplevel"], git_dir)
  File "/opt/conda/lib/python3.10/site-packages/foundry_dev_tools/config.py", line 341, in execute_as_subprocess
    return subprocess.run(
  File "/opt/conda/lib/python3.10/subprocess.py", line 503, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/opt/conda/lib/python3.10/subprocess.py", line 971, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/opt/conda/lib/python3.10/subprocess.py", line 1847, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'git'
```

# Checklist

### DIFF
--- a/src/foundry_dev_tools/config.py
+++ b/src/foundry_dev_tools/config.py
@@ -367,4 +367,7 @@ def _find_project_config_file(project_directory: Path) -> str:
         raise ValueError()
     # FileNotFoundError is thrown when git is not installed
     except (subprocess.CalledProcessError, FileNotFoundError) as exc:
+        warnings.warn(
+            "Could not find top-level directory of project, is git not installed?"
+        )
         raise ValueError from exc
--- a/src/foundry_dev_tools/config.py
+++ b/src/foundry_dev_tools/config.py
@@ -365,5 +365,6 @@ def _find_project_config_file(project_directory: Path) -> str:
                 )
                 return foundry_local_project_config_file
         raise ValueError()
-    except subprocess.CalledProcessError as exc:
+    # FileNotFoundError is thrown when git is not installed
+    except (subprocess.CalledProcessError, FileNotFoundError) as exc:
         raise ValueError from exc
--- a/src/foundry_dev_tools/config.py
+++ b/src/foundry_dev_tools/config.py
@@ -365,9 +365,13 @@ def _find_project_config_file(project_directory: Path) -> str:
                 )
                 return foundry_local_project_config_file
         raise ValueError()
-    # FileNotFoundError is thrown when git is not installed
-    except (subprocess.CalledProcessError, FileNotFoundError) as exc:
-        warnings.warn(
-            "Could not find top-level directory of project, is git not installed?"
+    except FileNotFoundError as exc:
+        LOGGER.debug(
+            "Project-based config file could not be loaded due to missing git installation"
+        )
+        raise ValueError from exc
+    except subprocess.CalledProcessError as exc:
+        LOGGER.debug(
+            "Project-based config file could not be loaded, is project not managed with git?"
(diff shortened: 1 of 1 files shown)


jonas-w:
Thanks, looks good!

I think it should print a warning if git is not installed, that people know why their project specific config files won't get loaded.

bernhardschaefer:
Good point. I added a warning, let me know if you want me to rephrase it.

jonas-w:
Thanks for the quick reply, would it be possible to print the warning only on a FileNotFoundError and only if the filename of FileNotFoundError matches git. (IIRC FileNotFoundError has filename attribute)

Otherwise (i think?) if the directory was not found, which may be the case when you don't work with git, the warning would also be thrown, even if git was installed. 

And I would rephrase the warning to explicitly say, that if git is not installed project specific configs can't be loaded, something along these lines.

nicornk:
I would favor to DEBUG log instead of warn. Most users are not using the project config, so I would not confuse them with the warning.
Is this the only place where git is used, or do we need to add the new catched Exception somewhere else as well?

jonas-w:
@nicornk you are probably right, too much noise, sorry didn't think about that. In _dataset.py we use git to get the branch of the dataset, but the error is caught and it falls back to master, that's the only other occurence of git.
https://github.com/emdgroup/foundry-dev-tools/blob/284bccad8783f00af3bad5026b24f17849152e8b/src/transforms/api/_dataset.py#L243-L261

nicornk:
@bernhardschaefer Please fix the imports for the exceptions and remove the warning again. Thanks!

This PR needs to be merged and released to get our build green here 
https://github.com/conda-forge/staged-recipes/pull/22324

bernhardschaefer:
@nicornk I changed warnings to logging as requested
@jonas-w I created separate messages for both exceptions as requested. If you want I can also check for the filename attribute, but the question is what to log if the filename is not equal to git. To do this properly I would prefer to explicitly check if git is installed at the beginning of the function (e.g. with  `has_git = shutil.which("git") is not None`) and get rid of the overarching try catch. In order not to delay the conda-forge release this could also be done in a follow-up PR, though.

jonas-w:
Thanks @bernhardschaefer for your contribution. Sorry for the pressure!

I'll check your other solution out, but i merged it as is, so we won't block the conda-forge recipe.

PR #8, opened 2023-03-23
Use gatekeeper:view-resource over compass:view

Dgleish:
# Summary

[check_read_access](https://github.com/emdgroup/foundry-dev-tools/blob/10b7e748c9ab1e4230fdd290cedc67bcd5737843/src/foundry_dev_tools/foundry_api_client.py#L932) currently uses `compass:view` to check if the user has read access to the dataset. `compass:view` is not treated as a single operation but the Viewer Role, which translates to many operations. Unfortunately, in some cases the Viewer Role will not map to `compass:view` but something else. Therefore it's possible to have the Viewer role on a dataset but not `compass:view`.

Instead `compass:read-resource` is a better operation to be checked. Worth noting that because of markings / custom Roles it can still be possible that someone with `compass:read-resource` doesn't have access to the data. There isn't any single operation you can check to guarantee data access.

# Checklist

- [X] Included tests (or is not applicable).

### DIFF
--- a/src/foundry_dev_tools/foundry_api_client.py
+++ b/src/foundry_dev_tools/foundry_api_client.py
@@ -915,7 +915,7 @@ class FoundryRestClient:
             dataset_path_or_rid (str): Path to dataset (e.g. /Global/...)
                 or rid of dataset (e.g. ri.foundry.main.dataset...)
             branch (str): branch of the dataset
-            check_read_access (bool): default is True, checks if the user has read access ('compass:read-resource')
+            check_read_access (bool): default is True, checks if the user has read access ('gatekeeper:view-resource')
                 to the dataset otherwise exception is thrown
 
         Returns:
@@ -930,7 +930,7 @@ class FoundryRestClient:
         dataset_rid = dataset_details["rid"]
         dataset_path = dataset_details["path"]
         if check_read_access:
-            if "compass:read-resource" not in dataset_details["operations"]:
+            if "gatekeeper:view-resource" not in dataset_details["operations"]:
                 raise DatasetNoReadAccessError(dataset_rid)
         return {
             "dataset_path": dataset_path,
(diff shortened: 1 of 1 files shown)


jonas-w:
Did you have issues with the first approach by only checking 'compass:view', or why do you want to change this? Because 'compass:read-resource' seems not to be the correct way to check this.

fmsf:
We had a person with owner access to a project, could see and manipulate the dataset data but didn't had `compass:view` this was throwing a `No read access to dataset`. When you have custom role sets viewer doesn’t correspond to compass:view 

`compass:read-resource` is a more correct operation that should work on both cases

jonas-w:
@fmsf okay thanks for the clarification, on our Instance this seems to bring up many many false positives, that's why i thought it would be incorrect. Is this the only universal way that could work?

Dgleish:
Ah sorry `compass:read-resource` comes with Discoverer and so `gatekeeper:view-resource` is probably going to be a better shout. Will update that.

nicornk:
Thanks for the contribution @Dgleish! Just verified on our stack that gatekeeper:view-resource is present as well.

Some background information why `get_dataset_identity` contains this manual check of the (read) operations...
In our apps we use the following helper function a lot - this function caches SQL Queries to foundry and the cache is only invalidated if there exists a new transaction on the source dataset. We need to make sure that the user has access to the dataset, that's why the operations check was added to the `get_dataset_identity` function.

```

def query_with_cache(query: str) -> pd.DataFrame:
    foundry_client = FoundryRestClient()

    @st.cache_data
    def query_cached(query: str, last_transaction_rid: str):
        """
        Cached Query against FoundrySQL Endpoint
        :param query: SQL Query to FoundrySQL
        :param last_transaction_rid: last_transaction_rid is used for cache invalidation
        :return: (pd.DataFrame) Result of the Query
        """
        print(f"{query=} {last_transaction_rid=}")
        return foundry_client.query_foundry_sql(query=query)

    dataset_path = query.split("`")[1]
    last_transaction_rid = foundry_client.get_dataset_identity(dataset_path)[
        "last_transaction_rid"
    ]
    return query_cached(query=query, last_transaction_rid=last_transaction_rid)
```

PR #15, opened 2023-04-18
Code Refactor with ruff instead of pylint

jonas-w:
# Summary
- switched from pylint to ruff
- removed pylint
- activated many more rules than pylint could offer
- pre-commit hook with auto fix enabled
- activated isort in ruff
- use black instead of ufmt+usort
- used Pathlib more often
- converted setup.cfg to pyproject.toml
  - renamed testing extras
- started linting tests folder, which was excluded
- experimented with mypy
- moved config initialization to config.py <- this removed some problematic circular imports
- raise_for_status_verbose prints now to stderr instead of stdout
- fixed a bug, that environment variables didn't take precedence over config files
- set requests timeout to 60
- replaced the subprocess calls with immitations of the git commands used

# Checklist


pyproject.toml
```diff
@@ -2,21 +2,119 @@
...
+[project.urls]
+Homepage = "https://emdgroup.github.io/foundry-dev-tools"
+Documentation = "https://emdgroup.github.io/foundry-dev-tools"
+Source = "https://github.com/emdgroup/foundry-dev-tools"
+Tracker = "https://github.com/emdgroup/foundry-dev-tools/issues"
+Changelog = "https://emdgroup.github.io/foundry-dev-tools/changelog.html"
+
+[tool.setuptools]
+zip-safe = false
+# packages = ["foundry-dev-tools", "transforms"]
```
nicornk: Remove uncommented? Is this still required? Did you manually check the resulting package, can it be installed?

src/foundry_dev_tools/__init__.py
```diff
@@ -1,19 +1,19 @@
 """Foundry Clients developed at Merck KGaA, Darmstadt, Germany."""
-from transforms import __version__
+from importlib.metadata import PackageNotFoundError, version
```
nicornk, line 2: Can you doublecheck if this is available in the standard library starting python 3.8?

src/foundry_dev_tools/foundry_api_client.py
```diff
@@ -43,20 +49,27 @@
 
 LOGGER = logging.getLogger(__name__)
 
+# duplicate code from __init__.py
+# if we import __version__ from __init__.py
+# we will have a circular import
+# which may have unintended side effects
+try:
+    FDT_VERSION = version(__name__)
```
nicornk, line 57: What is the practical consequence of this?

src/foundry_dev_tools/foundry_api_client.py
```diff
@@ -1056,31 +1083,23 @@ def download_dataset_file(
...
         resp = self._download_dataset_file(
             dataset_rid, view, foundry_file_path, stream=True
         )
-        if path.isdir(local_file_path):
-            resp.close()
-            return ""
-        with open(local_file_path, "wb") as out_file:
-            if resp.status_code != 200:
+        with local_path.open(mode="wb") as out_file:
+            if resp.status_code != requests.codes.ok:
```
nicornk: Should we just use our regular raise_for_status here instead of the custom check?

src/foundry_dev_tools/foundry_api_client.py
```diff
@@ -2070,38 +2081,28 @@ def _get_oauth2_client_credentials_token(
...
+    )
 
     return credentials.token
 
 
 def _request(*args, **kwargs):
-    kwargs["timeout"] = 599
-    # pylint: disable=missing-timeout
-    return requests.request(*args, **kwargs)
+    return requests.request(*args, **kwargs, timeout=599)
```
nicornk: What if someone passes timeout explicitly?

src/foundry_dev_tools/utils/caches/spark_caches.py
```diff
@@ -201,16 +199,19 @@ def _cleanup_old_transactions(self, dataset_identity: dict):
...
-        assert (
+        if (
             all_transactions_sorted[0].rsplit(".", 1)[0]
-            == dataset_identity["last_transaction_rid"]
-        ), "Cache dir not in sync with db."
+            != dataset_identity["last_transaction_rid"]
+        ):
+            print(all_transactions_sorted[0], file=sys.stderr)
+            print(dataset_identity["last_transaction_rid"], file=sys.stderr)
+            raise RuntimeError("Cache dir not in sync with db.")
```
nicornk: Please delete cache at ~/path-to-cache and restart compute.

nicornk (approved):
only minor comments

src/foundry_dev_tools/__init__.py
jonas-w: https://docs.python.org/3.8/library/importlib.metadata.html

src/foundry_dev_tools/foundry_api_client.py
jonas-w: as mentioned above there will be circular imports.

The problem is, we import foundry_api_client in the __init__.py, so the foundry_dev_tools module is not fully initialized.

That is why we also shouldn't use "from foundry_dev_tools import FoundryRestClient,..." in our own code, but "from foundry_dev_tools.foundry_api_client import FoundryRestClient"

src/foundry_dev_tools/config.py
```diff
@@ -320,7 +320,10 @@ def get_config(self, overwrite_config: "dict | None" = None):
...
 
 def _traverse_to_git_project_top_level_dir(git_dir: Path) -> Path:
-    return Path(execute_as_subprocess(["git", "rev-parse", "--show-toplevel"], git_dir))
+    for p in git_dir.resolve().parents:
+        if p.joinpath(".git").is_dir():
+            return p
+    return None
 
 
 def execute_as_subprocess(args: list, cwd: Path) -> str:
```
nicornk: @jonas-w Can we remove this now that its not used anymore?

PR #27, opened 2023-07-01
Implement foundry view support 

nicornk:
Implement foundry view support by using SQL Server "SELECT * Query" to retrieve content of view.

Limitation: We can't check the size of the view upfront and therefore the configured config thresholds are not taking effect. (Should we let this query timeout with a short timeout?)

* return 'last_transaction' from api.get_dataset_identity because it contains 'view' (bool) in 'records' key as well as dataset statistics.
* use statistics from 'last_transaction' saving one API call to stats endpoint
* CachedFoundryClient fetch_dataset/load_dataset can now load views, added unit test for this logic.
* Adopted tests and mocks.

# Summary

# Checklist


nicornk:
@jonas-w  can you move the is_view logic into a re-usable function please?

Issue #45, opened 2023-12-18
[Feature Request]: implement transforms.api.lightweight


bernhardschaefer:
### Issue checklist

- [X] This is a **feature request**/**enhancement**. And not a bug.
- [X] I searched through the GitHub issues and this feature/enhancement has not been requested before.
- [X] I have installed the latest version of Foundry DevTools and don't use an unsupported python version.
- [X] Others could also benefit from this feature or enhancement and it is **not** a very specific use case.

### Feature use case

The Foundry Transforms Python API contains a relatively new [@lightweight](https://www.palantir.com/docs/foundry/transforms-python/transforms-python-api/#lightweight) decorator.
At the moment I cannot use lightweight transforms locally as foundry dev tools does not yet implement the decorator.

### Description of the Feature

With the added feature, I could run lightweight transforms locally.

Example transform (copied from the [Foundry docs](https://palantir.mcloud.merckgroup.com/docs/foundry/transforms-python/lightweight-api/#data-formats)):
```python
@lightweight
@transform(output=Output('/Project/folder/output'), dataset=Input('/Project/folder/input'))
def compute(output, dataset):
    polars_df = dataset.polars()         # polars_df is a polars.DataFrame object
    lazy_df = dataset.polars(lazy=True)  # activate streaming mode, lazy_df is a polars.LazyFrame
    pandas_df = dataset.pandas()         # pandas_df is a pandas.DataFrame
    arrow_table = dataset.arrow()        # arrow_table is a pyarrow.Table
    out.write_table(lazy_df)             # either of the above can be passed to write_table
```

Local execution:
```python
from myproject.datasets import my_transform
df = my_transform.compute.compute()
```

PR #47, opened 2024-01-23
Add preview support for Lightweight transforms

schmelczer:
# Summary

Address https://github.com/emdgroup/foundry-dev-tools/issues/45

Add support for running local preview for [`@lightweight` and `@transform_polars` Transforms](https://www.palantir.com/docs/foundry/transforms-python/lightweight-api/) which [were recently added](https://www.palantir.com/docs/foundry/announcements/2023-11/#introducing-faster-transforms-for-small-to-medium-sized-datasets) to the Foundry Transforms library.

- The main contribution is adding local mock implementations for `@lightweight`, `@transforms_polars`, `LightweightTransformInput`, and `LightweightTransformOutput`. 
- However, next to this, this PR also proposes to make `Input` download the datasets and load the Spark DataFrames lazily. With this, we can avoid initialising Spark when using Lightweight transforms, which shaves off a few seconds from every preview run. 
  > This also has the added benefit of improving the initial startup overhead in larger repositories.
- A new method, `get_local_path_to_dataset`, is added to `Input` to lazily load the underlying files of the Dataset without having to initialise Spark when running Lightweight transforms.

## Limitations

- Bring-your-own-container workflows and setting resource limits (as-well-as provisioning GPUs) using `@lightweight` is not yet supported when running preview locally.
- When running `fdt build`, the build logs are fetched from spark reporter which is not available in the case of (lightweight) container transform. We could use foundry-telemetry-service's readV3 endpoint to tail logs, however, that has an at least 1 minute latency. Fortunately, a new, real-time log fetching endpoint is soont to be released, so we can transition to using that in a follow-up PR.
- CSV datasets are supported for Lightweight, but the Foundry schema is not loaded [(just like for regular transforms)](https://emdgroup.github.io/foundry-dev-tools/architecture.html#known-limitations-contributions-welcome)

# Checklist


schmelczer:
I'm coordinating with legal regarding the CLA

jonas-w:
@schmelczer thank you so much for this great contribution, I'll take a closer look at it tomorrow!

src/transforms/api/_decorators.py
```diff
@@ -5,11 +5,147 @@
...
+    memory_mb: Optional[float] = None,
+    memory_gb: Optional[float] = None,
+    gpu_type: Optional[str] = None,
+    container_image: Optional[str] = None,
+    container_tag: Optional[str] = None,
+    container_shell_command: Optional[str] = None,
+):
+    """Turn a transform into a Lightweight transform.
+
+    In order to use this decorator, `foundry-transforms-lib-python` must be added as a dependency.
```
jonas-w: Minor nit: The `foundry-transforms-lib-python` is not needed with Foundry DevTools

src/transforms/api/_decorators.py
```diff
@@ -5,11 +5,147 @@
...
+    and more cost-effective for small to medium-sized datasets. Lightweight transforms also provide more methods for
+    accessing datasets; however, they only support a subset of the API of a regular transforms, including Pandas and
+    the filesystem API. For more information, see the public documentation available at
+    https://www.palantir.com/docs/foundry/transforms-python/lightweight-overview
+
+    Args:
+        cpu_cores (float, optional): The number of CPU cores to request for the transform's container,
+            can be a fraction.
+        memory_mb (float, optional): The amount of memory to request for the container, in MB.
+        memory_gb (float, optional): The amount of memory to request for the container, in GB.
```
jonas-w: Could you also document, already in the docstring, that these won't take effect locally with Foundry DevTools?

src/transforms/api/_decorators.py
```diff
@@ -5,11 +5,147 @@
...
+        user code is given and is expected to return Polars DataFrames.
+
+        Note that spark profiles can't be used with lightweight transforms, hence, neither with @transforms_polars.
+
+        >>> @transform_polars(
+        ...     Output('ri.main.foundry.dataset.out'),  # An unnamed Output spec
+        ...     first_input=Input('ri.main.foundry.dataset.in1'),
+        ...     second_input=Input('ri.main.foundry.dataset.in2'),
+        ... )
+        ... def my_compute_function(ctx, first_input, second_input):
```
jonas-w: `ctx` is not supported in lightweight transforms

src/transforms/api/_transform.py
```diff
@@ -226,6 +280,95 @@ def filesystem(self):
...
+
+    def pandas(self) -> "pandas.DataFrame":
+        """A Pandas DataFrame containing the full view of the dataset."""
+        import pandas as pd
+
+        return (
+            pd.concat(map(pd.read_parquet, self._parquet_files), ignore_index=True)
+            if self._parquet_files
+            else pd.concat(
+                map(pd.read_csv, Path(self._path).glob("*.csv")), ignore_index=True
```
jonas-w: use `self._csv_files` instead

src/transforms/api/_transform.py
```diff
@@ -226,6 +280,95 @@ def filesystem(self):
...
+        """Download the dataset's underlying files and return a Path to them."""
+        return self._path
+
+    def pandas(self) -> "pandas.DataFrame":
+        """A Pandas DataFrame containing the full view of the dataset."""
+        import pandas as pd
+
+        return (
+            pd.concat(map(pd.read_parquet, self._parquet_files), ignore_index=True)
+            if self._parquet_files
```
jonas-w, line 311: Use walrus operator, so it searches for parquet files only once.

EDIT: I see that both methods get used multiple times, so switching _csv_files/_parquet_files to `@cached_property` could improve performance with big datasets.

```suggestion
            pd.concat(map(pd.read_parquet, parquet_files), ignore_index=True)
            if parquet_files := self._parquet_files
```

schmelczer:
We're still actively looking into the CLA

schmelczer:
The CLA has been cleared, so I've signed it! Feel free to take a look at the PR @jonas-w, I'm happy to discuss the details, especially if there're any counter-arguments against the lazy dataset loading

jonas-w (changes requested):
Sorry I just realized that I need to "submit" my review. This should hopefully make the comments visible that I posted last week!

nicornk:
Thanks for implementing the laziness in the decorators, something that was on our long list as well.

@schmelczer 🙏 thanks for the contribution.

src/transforms/api/_transform.py
```diff
@@ -226,6 +280,95 @@ def filesystem(self):
...
 
 
+class LightweightTransformInput(TransformInput):
+    """LightweightTransformInput class, passed when using @lightweight decorator."""
+
+    def __init__(self, input_arg: Input):
+        super().__init__(input_arg)
+        self.branch = input_arg.branch
+        self.rid = input_arg.get_dataset_identity()["dataset_rid"]
+        self._path = input_arg.get_local_path_to_dataset()
```
nicornk: Can you move `self._path` up to parent class`TransformInput` and use it in when the `FileSystem` (line 275) is constructed?  Also rename `_path` to `_local_path` to avoid confusion with Compass Path in Foundry (dataset_path).

src/transforms/api/_dataset.py
```diff
@@ -135,15 +142,13 @@ def _online(
             branch,
         )
         self._cached_client._fetch_dataset(dataset_identity, branch)
```
nicornk: This line needs to be deleted to make this lazy for file-only datasets as well.

src/transforms/api/_transform.py
```diff
@@ -356,4 +553,5 @@ def open(self, path, mode="w", **kwargs):  # noqa: A003
             IO:
 
         """
+        self._fs.makedirs(os.path.dirname(path), recreate=True)
```
nicornk, line 553: why is this required now?

nicornk (changes requested):
- Please double check the lazyness on the file-only (no-schema) datasets (see review comments)
- `fdt build`: Is there an easy way for us to detect if the build is using `@lighweight` and exit out gracefully until we support tailing logs with the "real-time log fetching endpoint"?  My assumption is, that at the moment it will probably timeout or throw an error.

Once we have a arrow-based local cache implementation @lighweight transforms should run completely without spark locally as well :)

src/transforms/api/_decorators.py
schmelczer: Good point!

src/transforms/api/_transform.py
schmelczer: Great idea!

src/transforms/api/_transform.py
schmelczer: Excellent idea, this is a neat refactor!

src/transforms/api/_transform.py
schmelczer: My bad, I should've highlighted it, it's an unrelated fix.

If I take this transform:
```py
@lightweight
@transform(
    my_input=Input("ri.foundry.main.dataset.888a07e8-1d07-427a-9bc4-b4dc1c40708a"),
    my_output=Output('ri.foundry.main.dataset.2375002f-8ee0-4ceb-a182-f573eb80f8f3')
)
def copy_files(my_input, my_output):
    for file in my_input.filesystem().ls():
        with my_input.filesystem().open(file.path, "rb") as f1:
            with my_output.filesystem().open(file.path, "wb") as f2:
                f2.write(f1.read())
```

When running the Preview with fdt, I would get:
```
(base) ➜  Lightweight-examples git:(master) ✗ /Users/aschmelczer/miniconda3/bin/python /Volumes/git/Lightweight-examples/transforms-python/src/myproject/datasets/api.py
/Volumes/git/foundry-dev-tools/src/transforms/api/_decorators.py:80: UserWarning: Setting resources for @lightweight transforms will have no effect when running locally.
  warnings.warn(
Traceback (most recent call last):
  File "/Volumes/git/Lightweight-examples/transforms-python/src/myproject/datasets/api.py", line 44, in <module>
    print(copy_files.compute())
          ^^^^^^^^^^^^^^^^^^^^
  File "/Volumes/git/foundry-dev-tools/src/transforms/api/_transform.py", line 97, in compute
    return handlers[self._type]()
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/Volumes/git/foundry-dev-tools/src/transforms/api/_transform.py", line 178, in _compute_lightweight
    self(**kwargs)
  File "/Volumes/git/foundry-dev-tools/src/transforms/api/_transform.py", line 81, in __call__
    return self._compute_func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Volumes/git/Lightweight-examples/transforms-python/src/myproject/datasets/api.py", line 17, in copy_files
    with my_output.filesystem().open(file.path, "wb") as f2:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Volumes/git/foundry-dev-tools/src/transforms/api/_transform.py", line 552, in open
    return self._fs.open(path, mode, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/aschmelczer/miniconda3/lib/python3.11/site-packages/fs/base.py", line 1228, in open
    bin_file = self.openbin(path, mode=bin_mode, buffering=buffering)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/aschmelczer/miniconda3/lib/python3.11/site-packages/fs/memoryfs.py", line 513, in openbin
    raise errors.ResourceNotFound(path)
fs.errors.ResourceNotFound: resource 'spark/part-00002-4e7e654b-c712-4a0f-afdd-d7e89e4425e8-c000.snappy.parquet' not found
```

However, in Foundry, this is allowed and builds successfully (both for Lightweight and regular transforms).

schmelczer:
Thanks for giving this a proper review, I've addressed your comments! 

I've done some further testing with schema-less datasets for both lightweight & regular transforms and preview seems to be working as expected! As for the logging, there isn't a trivial way of implementing this (we could look at the JobSpec though) but the foundry telemetry service update is just around the corner which will work with both Spark and Lightweight transforms!

src/transforms/api/_transform.py
nicornk: Makes sense, can you modify the test `test_transform_one_input`? I think the line `output1.filesystem()._fs.makedir("subfolder")` can now be removed and the test should still stay green.

src/transforms/api/_transform.py
```diff
@@ -218,10 +273,95 @@ def filesystem(self):
...
+    @cached_property
+    def _csv_files(self) -> List[Path]:
+        return list(Path(self._local_path).glob("**/*.csv"))
+
+    def dataframe(self):
+        raise NotImplementedError(
+            "Lightweight transforms do not support the dataframe method."
+        )
+
+    def path(self) -> str:
```
nicornk, line 301: Is `LightweightTransformInput` changing the API spec of `TransformInput` where `path` is defined as "The Project path of the input dataset."
If yes, we can keep the code as it is here.

https://www.palantir.com/docs/foundry/transforms-python/transforms-python-api-classes/#transforminput

src/transforms/api/_transform.py
```diff
@@ -186,8 +240,9 @@ class TransformInput:
     """TransformInput class, passed when using @transform decorator."""
 
     def __init__(self, input_arg: Input):
-        self._dataframe = input_arg.dataframe()
+        self._input_arg = input_arg
         self._dataset_identity = input_arg.get_dataset_identity()
+        self._local_path = input_arg.get_local_path_to_dataset()
```
nicornk: I thought about this again and right now it will trigger a download already in the constructor here, no-matter if it's used or not. 
Maybe you can directly call `self._input_arg.get_local_path_to_dataset()` in the filesystem() function and not store `self._local_path` in `TransformInput`

src/transforms/api/_transform.py
```diff
@@ -218,10 +273,95 @@ def filesystem(self):
             FileSystem:
                 A `FileSystem` object for reading from `Foundry`.
         """
-        return FileSystem(
-            base_path=get_dataset_path(
-                foundry_dev_tools.config.Configuration["cache_dir"],
-                self._dataset_identity,
+        return FileSystem(base_path=self._local_path)
```
nicornk: `return FileSystem(base_path=self._input_arg.get_local_path_to_dataset())`

src/transforms/api/_transform.py
```diff
@@ -218,10 +273,95 @@ def filesystem(self):
...
+
+
+class LightweightTransformInput(TransformInput):
+    """LightweightTransformInput class, passed when using @lightweight decorator."""
+
+    def __init__(self, input_arg: Input):
+        super().__init__(input_arg)
+        self.branch = input_arg.branch
```
nicornk, line 283: add `self._input_arg.get_local_path_to_dataset()` here.
Without the call a lightweight transformInput does not make sense anyway, so good to call here.

nicornk (approved):
Thanks @schmelczer , just two minor changes before we can merge and release this.
@jonas-w After merge, could you please prepare the changelog and release a new version?

tests/test_transforms.py
```diff
@@ -102,6 +104,22 @@ def return_df(one, dataset_path, branch):
...
+    destination = os.sep.join(
+        [
+            foundry_dev_tools.config.Configuration[
+                "cache_dir"
+            ],  # this is a tmp dir in unit tests
+            identity["dataset_rid"],
+            identity["last_transaction_rid"] + ".parquet",
+        ]
+    )
+    df.write.parquet(os.path.join(destination, "spark"))
```
nicornk: @schmelczer No idea why but I had to add `mode='overwrite'` to make the tests go through in my codespace
`df.write.parquet(os.path.join(destination, "spark"), mode='overwrite')`

src/transforms/api/_transform.py
schmelczer: Yes, we should update the public docs to include the APIs of `LightweightTransformInput` and output, I'll add it to our todo list

src/transforms/api/_transform.py
schmelczer: Yes, it does pass now without that line!

src/transforms/api/_transform.py
nicornk: Just wanted to edit that I now saw the new public docs about input.path() here:

https://www.palantir.com/docs/foundry/transforms-python/lightweight-examples/#working-with-apache-datafusion-and-duckdb

In this case the code here is fine. (Although I don't like that you re-purposed the `path` property ;) )

src/transforms/api/_transform.py
schmelczer: We also use `self._local_path` in `LightweightTransformInput` which is a subclass. What would you say if we turned `_local_path` into a private property?

nicornk:
lgtm

jonas-w:
@schmelczer would it be possible to remove this merge commit [`d111123` (#47)](https://github.com/emdgroup/foundry-dev-tools/pull/47/commits/d111123d363733bcc5c42fa55089e51c77acb71e) and instead do a rebase? This way the history commit history would stay a bit cleaner after we merge it into main.

Otherwise it looks good to me!

schmelczer:
Sure, I've dropped the merge commit @jonas-w!

jonas-w (approved):

jonas-w:
Should be released by now https://pypi.org/project/foundry-dev-tools/

Thanks again @schmelczer!

PR #49, opened 2024-02-01
Improve Contribution guide by introducing "complete" profile & clarifying java version

nicornk:
# Summary
While trying to run the unit tests on a new computer/within Codespaces I realized that the contribution guide could be improved. With these changes it should be straight forward to follow to get Green Unit Tests on the first try.

# Checklist

### DIFF
--- a/docs/develop.md
+++ b/docs/develop.md
@@ -14,9 +14,19 @@ pip install -e .
 This command will link the local version of this package into your pip package folder.
 Every change you make in the code is instantly applied.
 
+## Setup Development Environment
+
+The easiest option is to open the repository with GitHub Codespaces, which offers a free tier as well.
+Spark in combination with pyarrow currently works only with Java<21 (`sun.misc.Unsafe or java.nio.DirectByteBuffer.<init>(long, int) not available`). 
+In Codespaces you can change the Java Version to 17 using:
+
+```shell
+sdk default java 17.0.9-ms
+```
+
 ## Run unit tests
 
-Install the test dependencies and execute pytest.
+Install all test dependencies and execute pytest.
 
 ```shell
 pip install -e ".[complete]"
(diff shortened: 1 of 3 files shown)


jonas-w:
Maybe we should also add this to docs/installation.md and include a way to do this with mamba/conda via `mamba install 'openjdk==17.*' pyspark`? Or we could create a seperate documentation were we include all the information about pyspark, so we can link from docs/installation.md and docs/develop.md?

jonas-w:
The "complete" profile is a good idea, but what do you think about if we do an "all" extra, that contains transforms, cli and s3. Whiche end-users could use. And then a "dev" extra, which contains "all" + testing and integration + maybe pre-commit for contributors?

nicornk:
Good points. I feel like we overcomplicate a little bit. 
What if we include the dependencies of "s3" and "cli" in the default dependencies and than have "all" (default + transforms) and dev (all, integration, testing, pre-commit, tox)?

I think a seperate documentation about the pyspark and Java quirks make sense.

jonas-w:
> Good points. I feel like we overcomplicate a little bit. What if we include the dependencies of "s3" and "cli" in the default dependencies and than have "all" (default + transforms) and dev (all, integration, testing, pre-commit, tox)?
> 
> I think a seperate documentation about the pyspark and Java quirks make sense.

I think we should leave the extras as is for now, and think about it again in the v2, where we already thought about splitting the extras into packages. Though I wouldn't mind the extra "extras" 'all' and 'dev'.

I've tested a bit, and with `pip install -e '.'`, `pip freeze` shows 21 packages, but with `pip install -e '.[s3,cli]'`, it would be 48 packages, more than 2x amount of dependencies.

Making s3 the default, may be interesting when we start using the s3 compatible dataset API for other things.

We would also need to patch this again in the lambda layers, if we start depending on these packages per default 😅

jonas-w:
Obsolete as package management changed, so contribution guide also changed.

PR #52, opened 2024-03-12
v2

jonas-w:
# Summary

- Rewrite of Foundry DevTools v1
- v1 compability for FoundryRestClient and CachedFoundryClient
- extendable through entry points

# Checklist

# TODO

- [x] fix ruff errors
- [x] v1_to_v2_config tests
- [x] extra cli dependencies?
- [x] finish migrate config tests
- [x] migrate project config file
- [x] transforms as seperate package
- [x] change github publish workflow to use ~hatch~ pdm / make sure tests do run in PR.
- [ ] maybe provide more examples
- [x] check TODOs in code
- [ ] test v2 on more real transforms and v1 code
- [ ] see if the documentation workflow works
- [x] check if more tests are needed (especially multipass test is still missing)
- [x] discuss if we will use deperecation warnings for the v1 clients
- [x] test documentation example code snippets
- [x] polars s3 compat
- [ ] Blog Post / Announcement
- [x] TL;DR v1->v2 changes
- [x] git: document removing token from URL, maybe even a clone CLI
- [x] typo checks in docs
- [x] resolve PR conflicts
- [x] bump min python 3.10
- [x] remove transforms context singleton
- [x] which package to install in the docs/getting started guide? -transforms?

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -120,6 +120,7 @@ html_theme = "sphinx_book_theme"
 html_theme_options = {
     "show_navbar_depth": 2,
     "show_toc_level": 4,
+    "announcement":"Note: This is the documentation for Foundry DevTools v2",
 }
 
 # The name for this set of Sphinx documents.  If None, it defaults to


docs/configuration.md
```diff
@@ -0,0 +1,257 @@
+# Configuration
```
nicornk: I think we need a TLDR at the beginning of this file explaining how to make a config with one JWT and a hostname 

```

[credentials]
domain="palantir.mcloud.merckgroup.com"
token_provider.config.jwt="eyJwb..."
```
and one with oauth client_id/client_secret.
And than tell the user afterwards to run `fdt config` to check if all good.

src/foundry_dev_tools/__init__.py
```diff
@@ -1,14 +1,20 @@
 """Foundry Clients developed at Merck KGaA, Darmstadt, Germany."""
-from .__about__ import __version__
-from .cached_foundry_client import CachedFoundryClient
-from .config import FOUNDRY_DEV_TOOLS_DIRECTORY, INITIAL_CONFIG, Configuration
-from .foundry_api_client import FoundryRestClient
+
+from foundry_dev_tools.__about__ import __version__
```
nicornk: It seems the `__about__` file is only created after `pip install -e .` which makes running the tests from PyCharm impossible without installing the package first. That was better previously. Can you find a better solution that allows running tests without installing into the local environment first? Or convince me that this is now best practice in python packaging ;)

src/foundry_dev_tools/config/token_provider.py
```diff
@@ -0,0 +1,193 @@
...
+    "compass:view",
+    "compass:edit",
+    "compass:discover",
+    "api:write-data",
+    "api:read-data",
+    "build2:run-build-using-service",
+]
+
+
+class OAuthTokenProvider(CachedTokenProvider):
```
nicornk: Can you add tests for this class, testing the logical flow. E.g. if client_id && authorization code grant -> call to palantir oauth2 library, if client_id && client_secret && client credentials grant -> make http request etc.

src/foundry_dev_tools/utils/token_provider.py
```diff
@@ -1,137 +0,0 @@
-"""Token Provider for common UPTIMIZE AppService."""
```
nicornk: We have to leave a code path in (for compatibility reasons), where we check if os.environ['APP_SERVICE_TS'] is present and if yes we automatically try both token providers in this file.

src/foundry_dev_tools/utils/token_provider.py
```diff
@@ -1,137 +0,0 @@
...
-            except (
-                ImportError,
-                KeyError,
-                RuntimeError,
-                ModuleNotFoundError,
-                TypeError,
-            ):
-                return None
-
-    def get_streamlit_request_headers(self):
```
nicornk: this can be removed, we do not need to support it anymore

src/foundry_dev_tools/cli/git_credential_foundry.py
```diff
@@ -0,0 +1,66 @@
+"""A git credential helper.
+
+Can be used by adding:
+[credential "https://your-stack.palantirfoundry.com"]
+  helper =
```
nicornk: remove line?

src/foundry_dev_tools/cli/git_credential_foundry.py
```diff
@@ -0,0 +1,66 @@
+"""A git credential helper.
```
nicornk: Can you add docs for this feature? How exactly is the python code called from git?

src/foundry_dev_tools/resources/dataset.py
```diff
@@ -0,0 +1,620 @@
...
+            folder_path: the folder to upload
+            **kwargs: passed to :py:meth:`foundry_dev_tools.resources.dataset.Dataset.upload_files`
+
+        Returns:
+            the transaction used for uploading the files
+        """
+        kwargs.setdefault("transaction_type", api_types.FoundryTransaction.UPDATE)
+        return self.upload_files({str(f.relative_to(folder_path)): f for f in folder_path.rglob("*")}, **kwargs)
+
+    def save_model(
```
nicornk: I think we should leave it up to the user to pickle/dill/serialize the model and not just use pickle. 
Maybe we just delete this function? I also don't really like it on the CachedFoundryClient.

src/foundry_dev_tools/clients/data_proxy.py
```diff
@@ -0,0 +1,549 @@
...
+            for key, value in path_file_dict.items():
+                self.upload_dataset_file(
+                    dataset_rid,
+                    transaction_rid,
+                    path=value,
+                    path_in_foundry_dataset=key,
+                )
+        else:
+
+            def _upload_file(path_in_dataset: PathInDataset, path: Path):
```
nicornk: not used.

src/foundry_dev_tools/resources/dataset.py
```diff
@@ -0,0 +1,620 @@
...
+
+    import pandas as pd
+    import pandas.core.frame
+    import pyarrow as pa
+    import pyspark.sql
+
+    from foundry_dev_tools.config.context import FoundryContext
+
+
+class Dataset(resource.Resource):
```
nicornk: If you want to keep this in, this class needs a good end2end integration test, testing all functions.

src/foundry_dev_tools/utils/config.py
```diff
@@ -0,0 +1,216 @@
...
+        Path | None: Path to the project config or None if no file was found
+    """
+    git_directory = git_toplevel_dir(project_directory, use_git=use_git)
+    if not git_directory:
+        return None
+
+    return git_directory.joinpath(PROJECT_CFG_FILE_NAME)
+
+
+def get_environment_variable_config() -> dict:
```
nicornk: This method needs some tests

src/foundry_dev_tools/config/config.py
```diff
@@ -0,0 +1,189 @@
...
+        # the jwt token provider needs a config
+        if tp_config:
+            ci = check_init(JWTTokenProvider, "credentials", {"host": host, **tp_config})
+            return JWTTokenProvider(**ci)
+        msg = (
+            "To authenticate with Foundry you need a TokenProvider. The token provider can be configured either via the"
+            " configuration file or the token_provider FoundryContext parameter."
+        )
+        raise TokenProviderConfigError(msg)
+    raise MissingCredentialsConfigError
```
nicornk: if no JWT, no auth, and APP_SERVICE_TS is set, use legacy streamlit/dash provider.

src/foundry_dev_tools/cli/config.py
```diff
@@ -0,0 +1,90 @@
...
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+if sys.version_info < (3, 11):
+    pass
+else:
+    pass
+
+
```
nicornk: Add migrate method fdt config migrate, that
a) migrate to toml
b) rename old config.

src/foundry_dev_tools/__init__.py
jonas-w: @nicornk will test with commiting a stub file

docs/configuration.md
jonas-w: I'll add a TLDR to configuration.md

tests/test_multipass_tpa.py
```diff
@@ -1,277 +0,0 @@
-"""Enable enable_third_party_application
```
nicornk: Needs to be migrated into integration folder

src/foundry_dev_tools/utils/token_provider.py
jonas-w: @nicornk it would be better if we could set the token provider implementation directly instead of the extra code path to check the `APP_SERVICE_TS` env var.

We could set `FDT_CREDENTIALS__TOKEN_PROVIDER__NAME` which would use the already existing code path to get the token provider implementation

pyproject.toml
```diff
@@ -28,37 +28,39 @@ classifiers = [
...
   "pytest-asyncio",
   "requests-mock",
-  "fsspec",
-  "timeflake",
-  "polars"
+  "polars",
 ]
-transforms = ["pyspark>=3.0.0"]
+transforms = ["pyspark>=3.0.0", "fs"]
 cli = ["click", "inquirer", "websockets", "rich", "packaging"]
```
nicornk: Should we add those (cli) to the default dependencies?

src/foundry_dev_tools/__init__.py
jonas-w: it's not possible to ignore files in git that are already committed, so committing this file would mean everytime you installed the project and want to commit something you need to make sure to not commit the file.

Otherwise we could not use `hatch-vcs` and update the version numbers manually, and commit this file everytime there is a version update.

src/foundry_dev_tools/cli/git_credential_foundry.py
jonas-w: @nicornk the line is needed to force this helper for the foundry URL

https://git-scm.com/docs/gitcredentials/2.25.2#Documentation/gitcredentials.txt-helper

> If credential.helper is configured to the empty string, this resets the helper list to empty (so you may override a helper set by a lower-priority config file by configuring the empty-string helper, followed by whatever set of helpers you would like).

src/foundry_dev_tools/utils/token_provider.py
nicornk: We can't assume that any other env. variables apart from APP_SERVICE_TS are set, that indicate code is running on the app service.

src/foundry_dev_tools/resources/dataset.py
nicornk: Remove from dataset class and think about new design.

libs/foundry-dev-tools/src/foundry_dev_tools/clients/compass.py
```diff
@@ -416,3 +422,221 @@ def api_get_resources(
...
+        body = {"markingPatches": [{"markingId": marking_id, "patchOperation": str(path_operation_type)}]}
+
+        return self.api_request(
+            "POST",
+            f"markings/{rid}",
+            headers={"User-Bearer-Token": f"Bearer {user_bearer_token}"} if user_bearer_token else None,
+            json=body,
+            error_handling=ErrorHandlingConfig(
+                {
+                    "Compass:AutosaveResourceOperationForbidden": AutosaveResourceOperationForbiddenError,
```
jonas-w: put this also in the default error mapping, as this error type can be mapped directtly to that class for all requests, and remove this ErrorHandlingConfig as it won't be needed anymore.

libs/foundry-dev-tools/src/foundry_dev_tools/clients/compass.py
```diff
@@ -416,3 +422,221 @@ def api_get_resources(
...
+                        "disableInheritedPermissionsForPrincipals": [{
+                            "id": "..",
+                            "type": ".."
+                        }],
+                        "disableInheritedPermissions": False,
+                        "disableInheritedPermissionsType": ".."
+                    }
+                }
+             }
+        """
```
jonas-w: Method does not have a body, only a docstring

libs/foundry-dev-tools/src/foundry_dev_tools/clients/compass.py
```diff
@@ -416,3 +422,221 @@ def api_get_resources(
...
+
+        Args:
+            rid: resource identifier
+            **kwargs: gets passed to :py:meth:`APIClient.api_request`
+
+        Returns:
+            bool:
+                true if resource exists, false otherwise
+        """
+        result = self.api_resources_exist({rid}, kwargs).json()
```
jonas-w:
```diff
- result = self.api_resources_exist({rid}, kwargs).json()
+ result = self.api_resources_exist({rid}, **kwargs).json()
```

libs/foundry-dev-tools/src/foundry_dev_tools/clients/compass.py
```diff
@@ -416,3 +422,221 @@ def api_get_resources(
...
+
+        Returns:
+            dict:
+                with key-value pairs, where the key is the rid of the checked resource
+                and the value indicates whether the resource exists or not.
+        """
+        return self.api_request(
+            "POST",
+            "batch/resources/exist",
+            json=[*rids],
```
jonas-w:
```diff
- json=[*rids],
+ json=list(rids),
```

libs/foundry-dev-tools/src/foundry_dev_tools/clients/compass.py
```diff
@@ -618,25 +574,285 @@ def api_resources_exist(
...
+        """
+        list_len = len(rids)
+        if list_len < GET_PATHS_BATCH_SIZE:
+            batches = [rids]
+        else:
+            batches = [rids[i : i + GET_PATHS_BATCH_SIZE] for i in range(0, list_len, GET_PROJECTS_BATCH_SIZE)]
+
+        result: dict[api_types.FolderRid, dict[str, Any]] = {}
+        for batch in batches:
+            result = {**result, **self.api_get_projects_by_rids(batch).json()}
```
jonas-w: Unpacking dicts can be expensive, replace with `result.update(self.api_....)`

libs/foundry-dev-tools/src/foundry_dev_tools/clients/compass.py
```diff
@@ -618,25 +574,285 @@ def api_resources_exist(
...
+        """Returns a single project.
+
+        Args:
+            rid: resource identifier
+
+        Returns:
+            dict:
+                mapping of project attribute keys and their respective values
+        """
+        result = self.get_projects_by_rids([rid])
```
jonas-w: Could be replaced with the 'low level' API call, as you don't need the batch logic here.

libs/foundry-dev-tools/src/foundry_dev_tools/clients/compass.py
```diff
@@ -618,25 +574,285 @@ def api_resources_exist(
...
+
+        Returns:
+            response:
+                the response contains a json which is a dict with a list of projects
+                and an optional nextPageToken in case all the projects could not be fetched all at once
+        """
+        if decorations is not None:
+            decorations = get_decoration(decorations)
+        if page_size not in range(MINIMUM_PROJECTS_PAGE_SIZE, MAXIMUM_PROJECTS_PAGE_SIZE + 1):
+            page_size = DEFAULT_PROJECTS_PAGE_SIZE
```
jonas-w: Throw a warning that you are not using the number specified by the user.

libs/foundry-dev-tools/src/foundry_dev_tools/clients/compass.py
```diff
@@ -618,25 +574,285 @@ def api_resources_exist(
...
+            **kwargs: gets passed to :py:meth:`APIClient.api_request`
+
+        Returns:
+            response:
+                the response contains a json which is a dict with a list of projects
+                and an optional nextPageToken in case all the projects could not be fetched all at once
+        """
+        if decorations is not None:
+            decorations = get_decoration(decorations)
+        if page_size not in range(MINIMUM_PROJECTS_PAGE_SIZE, MAXIMUM_PROJECTS_PAGE_SIZE + 1):
```
jonas-w: Replace with `MINIMUM_PROJECTS_PAGE_SIZE <= page_size <= MAXIMUM_PROJECTS_PAGE_SIZE`

libs/foundry-dev-tools/src/foundry_dev_tools/clients/compass.py
```diff
@@ -427,7 +430,7 @@ def api_process_marking(
         self,
         rid: api_types.Rid,
         marking_id: api_types.MarkingId,
-        path_operation_type: api_types.PatchOperation | str,
```
jonas-w: What's the reason for this change?

libs/foundry-dev-tools/src/foundry_dev_tools/clients/compass.py
```diff
@@ -618,25 +574,285 @@ def api_resources_exist(
...
+        }
+
+        return self.api_request(
+            "POST",
+            "search/projects",
+            json=body,
+            **kwargs,
+        )
+
+    def search_projects(
```
jonas-w: When searching via this method I get the following error:

```
METHOD = POST
CONTENT-TYPE = application/json
ENDPOINT = /compass/api/search/projects
-------------------------------------------------------------------------------------------------------------------------------------------------PARAMETERS-------------------------------------------------------------------------------------------------------------------------------------------------
error_code = INVALID_ARGUMENT
error_name = Compass:InvalidSearchOffset
error_instance_id = 171afbc3-ef18-46d0-8b50-c8b94414f649
given_offset = 600
min_offset = 0
max_offset = 500
--------------------------------------------------------------------------------------------------------------------------------------------------RESPONSE--------------------------------------------------------------------------------------------------------------------------------------------------
ERROR_NAME = Compass:InvalidSearchOffset
ERROR_CODE = INVALID_ARGUMENT
ERROR_INSTANCE_ID = 171afbc3-ef18-46d0-8b50-c8b94414f649
STATUS = 400
```

Could you also add a check for the maximum offset in the lower level method and in this higher level method?

libs/foundry-dev-tools/src/foundry_dev_tools/clients/compass.py
```diff
@@ -618,25 +574,285 @@ def api_resources_exist(
...
+        rid: api_types.Rid,
+        grant_patches: set[api_types.RoleGrantPatch] | None = None,
+        disable_inherited_permissions_for_principals: set[api_types.UserGroupPrincipalPatch] | None = None,
+        disable_inherited_permissions: bool | None = None,
+        **kwargs,
+    ) -> requests.Response:
+        """Updates the role grants for a resource.
+
+        Args:
+            rid: resource identifier
```
jonas-w: Multiple instances of this in this file, but I'll use this as an example here.

Clarify what resource identifier is expected here.
In this case it is the resource identifier, for the resource for which roles will be updated.

libs/foundry-dev-tools/src/foundry_dev_tools/clients/compass.py
Silas2701: I actually wanted to use the former PatchOperation object in TypedDict for API models. But because the former PatchOperation was a enum class, the enum type would not resolve correctly in the dict (instead of 'ADD', the result was 'PatchOperation.ADD'). So I renamed it to PatchOperationType and introduced a new PatchOperation object of type Literal to be used in the API object classes extending TypedDict.
I didn't know if there was a more elegant way to solve it (and reduce redundancy as both PatchOperation and PatchOperationType basically represent the same).

Extract from [api_types.py](https://github.com/emdgroup/foundry-dev-tools/blob/v2/libs/foundry-dev-tools/src/foundry_dev_tools/utils/api_types.py):

```
PatchOperation = Literal["ADD", "REMOVE"]
"""Foundry PatchOperation enum."""


class PatchOperationType(str, Enum, metaclass=EnumContainsMeta):
    """Foundry patch operations for markings."""

    ADD = "ADD"
    """Add marking"""
    REMOVE = "REMOVE"
    """Remove marking"""

    def __str__(self) -> str:
        return self.value
```

jonas-w: Normally I'd also prefer to use just a Literal for this, but with Enums I can easily verify client side (without sending it to the API) if the input was correct, via this line, which you removed in your commit:

```python
# method accepts either the Enum or a string, if it is not a string, the input should already be correct
if isinstance(path_operation_type, str):
# we cast it to the Enum, which would throw an error if the string is not valid for this enum
    path_operation_type = api_types.PatchOperation(path_operation_type) 
```

I'll take a look at it again, and see if there is a better solution for this.
Typing in Python, sadly, isn't as easy as I wish it were.

libs/foundry-dev-tools/src/foundry_dev_tools/clients/compass.py
```diff
@@ -723,8 +731,24 @@ def api_search_projects(
         """
         if decorations is not None:
             decorations = get_decoration(decorations)
-        if page_size not in range(MINIMUM_PROJECTS_PAGE_SIZE, MAXIMUM_PROJECTS_PAGE_SIZE + 1):
-            page_size = DEFAULT_PROJECTS_PAGE_SIZE
+        if not MINIMUM_PROJECTS_PAGE_SIZE <= page_size <= MAXIMUM_PROJECTS_PAGE_SIZE:
+            page_size = (
+                MINIMUM_PROJECTS_PAGE_SIZE if page_size < MINIMUM_PROJECTS_PAGE_SIZE else MAXIMUM_PROJECTS_PAGE_SIZE
+            )
+            warnings.warn(f"Parameter `page_size` not within boundaries (1 <= N <= 500). Defaulting to {page_size}.")
```
jonas-w: Give the user a bit more information, what the page size was.

```python
        if page_size < MINIMUM_PROJECTS_PAGE_SIZE:
            warnings.warn(
                f"Parameter `page_size` ({page_size}) is less than the minimum page size ({MINIMUM_PROJECTS_PAGE_SIZE}). "
                f"Defaulting to {MINIMUM_PROJECTS_PAGE_SIZE}."
            )
            page_size = MINIMUM_PROJECTS_PAGE_SIZE
        elif page_size > MAXIMUM_PROJECTS_PAGE_SIZE:
            warnings.warn(
                f"Parameter `page_size` ({page_size}) is greater than the maximum page size ({MAXIMUM_PROJECTS_PAGE_SIZE}). "
                f"Defaulting to {MAXIMUM_PROJECTS_SEARCH_OFFSET}."
            )
```

libs/foundry-dev-tools/src/foundry_dev_tools/clients/compass.py
```diff
@@ -723,8 +731,24 @@ def api_search_projects(
...
+            and page_token.isdecimal()
+            and MINIMUM_PROJECTS_SEARCH_OFFSET <= int(page_token) <= MAXIMUM_PROJECTS_SEARCH_OFFSET
+        )
+
+        if not valid_page_token:
+            page_token = None
+            warnings.warn(
+                "Parameter `page_token` is invalid and will not be reset to default value. "
+                "For valid pattern see parameter description of this function."
+            )
```
jonas-w: The `page_token` is per default None, so this warning shows per default.

If you reset the `page_token` to None here, it would also refetch the first result page again in the end.

Replace the code with something like this:
```python
if page_token is not None and page_token.isdecimal():
   page_token_int = int(page_token)
   if (MINIMUM_PROJECTS_SEARCH_OFFSET > page_token_int or page_token_int > MAXIMUM_PROJECT_SEARCH_OFFSET):
    # raise an exception, which states what the page_token value is and what the problem is (similar to my comment above about the page_size changes).
```

Silas2701: You are right, I saw my mistake.

I would even exlude the `page_token.isdecimal()` into a separate condition, because we would otherwise miss the check for non-numeric values resulting in an error.

```
if page_token is not None:
    if page_token.isdecimal() is False:
        msg = (
            f"Parameter `page_token` ({page_token}) is expected to contain a number "
            f"as the starting offset for the request. "
            f"The search offset must be within the range from "
            f"{MINIMUM_PROJECTS_SEARCH_OFFSET} to {MAXIMUM_PROJECTS_SEARCH_OFFSET}."
        )
        raise ValueError(msg)

    page_token_int = int(page_token)
    if page_token_int < MINIMUM_PROJECTS_SEARCH_OFFSET:
        msg = (
            f"Parameter `page_token` ({page_token_int}) is less than "
            f"the minimum search offset ({MINIMUM_PROJECTS_SEARCH_OFFSET})"
        )
        raise ValueError(msg)
    if page_token_int > MAXIMUM_PROJECTS_SEARCH_OFFSET:
        msg = (
            f"Parameter `page_token` ({page_token_int}) is greater than "
            f"the maximum search offset ({MAXIMUM_PROJECTS_SEARCH_OFFSET})"
        )
        raise ValueError(msg)
```

libs/foundry-dev-tools/src/foundry_dev_tools/clients/compass.py
```diff
@@ -310,6 +310,25 @@ def get_paths(
             result = {**result, **self.api_get_paths(batch).json()}
         return result
 
+    def get_path(
+        self,
+        resource_id: api_types.Rid,
```
nicornk: `resource_rids`

Can you also refactor argument of method `get_paths`?

tests/integration/clients/test_compass.py
```diff
@@ -0,0 +1,608 @@
...
+    while len(grants) != expected_grant_size:
+        time.sleep(0.1)
+
+        result = TEST_SINGLETON.ctx.compass.get_resource_roles({rid})
+        grants = result["resourceRolesResultMap"][rid]["grants"]
+
+    return grants
+
+
+"""
```
nicornk: Please remove commented out code.

tests/integration/clients/test_compass.py
```diff
@@ -0,0 +1,608 @@
...
+    yield compass_folder_rid, compass_folder_name
+
+    # Delete test folder
+    delete_compass_folder(compass_folder_rid)
+
+
+@pytest.fixture()
+def marking_id():
+    """The id of `ls-use-case-foundry-devtools-dev-workspace` marking"""
+    return "3abebc75-4915-4b5c-8749-9ce06726b6e1"
```
nicornk: Please define as constant in `tests.integration.utils` and add a comment which permissions the user needs for the tests on the marking.

tests/integration/clients/test_compass.py
```diff
@@ -0,0 +1,608 @@
...
+
+
+@pytest.fixture()
+def test_project_rid():
+    """The rid of `ls-use-case-foundry-devtools-dev-workspace` project"""
+    return "ri.compass.main.folder.f94fc087-6068-4e66-ac27-978fce7b9d9d"
```
nicornk: I think you could update INTEGRATION_TEST_COMPASS_ROOT_RID with a new folder in the ls-use-case-foundry-devtools-dev-workspaceand re-use that.

Silas2701: @nicornk Sure, I can reset the INTEGRATION_TEST_COMPASS_ROOT_RID to some new folder in the ls-use-case-foundry-devtools-dev-workspace project.
But perhaps it would also be good, to have the project rid defined as constant in `tests.integration.utils` as it is explicitly required in some tests, e.g. to fetch a project by its rid or adding/removing a marking on project level.

Silas2701: And maybe it is even better to not update the INTEGRATION_TEST_COMPASS_ROOT_RID with the rid of some folder in the ls-use-case-foundry-devtools-dev-workspace. The current INTEGRATION_TEST_COMPASS_ROOT_RID is more "accessible" compared to a folder rid in the ls-use-case-foundry-devtools-dev-workspace project. Not everyone has access to the project and (whoever will work on the foundry dev tools in future) might not be able to execute the tests then. 
But if you think, that it is not such a big deal, I can update the INTEGRATION_TEST_COMPASS_ROOT_RID of course.

nicornk: Fine for me to leave the INTEGRATION_TEST_COMPASS_ROOT_RID and add another constant pointing to the new folder.

libs/foundry-dev-tools/src/foundry_dev_tools/cli/config.py
```diff
@@ -93,11 +93,17 @@ def _edit_file(f: Path):
 
 @click.group("config", invoke_without_command=True)
 @click.pass_context
-def config_cli(ctx: click.Context):
+@click.option(
+    "-p",
+    "--profile",
+    help=("The config profile to select.\n" "If not provided 'default' is used."),
```
jonas-w: @nicornk Why "default"?

After this update, I'd need to update my local config (valid before this commit) from:

```toml
[credentials]
domain = "..."
token_provider.config.jwt = "eyJ..."
```

to

```toml
[default.credentials]
domain = "..."
token_provider.config.jwt = "eyJ..."
```

Is this intended? Then I'd need to update the docs aswell.

Also setting the default in the CLI to "default", does not allow the user to have a `profile="profile name"` in their config to set the default profile.

libs/foundry-dev-tools/src/foundry_dev_tools/clients/compass.py
```diff
@@ -272,57 +272,57 @@ def api_get_path(self, rid: api_types.Rid, **kwargs) -> requests.Response:
             **kwargs,
         )
 
-    def api_get_paths(self, resource_ids: list[api_types.Rid], **kwargs) -> requests.Response:
+    def api_get_paths(self, resource_rids: list[api_types.Rid], **kwargs) -> requests.Response:
         """Get paths for RIDs.
 
         Args:
-            resource_ids: The identifiers of resources
```
jonas-w: `rid` already means 'resource identifier', after this change this variable name spelled out would be 'resource resource identifier', wouldn't it be better to remove the `resource_` prefix completely, if it's already clear that a `rid` is for a resource?

Silas2701: Should be clear from my understanding

libs/foundry-dev-tools/src/foundry_dev_tools/cli/config.py
nicornk: Hi @jonas-w ,

I was manually testing a config with two different stacks and different oauth configurations and used fdt config to verify the result. To be able to select the profile for testing I added the profile cli argument everywhere. I was not aware that the active profile can also be set in the config itself.

In the end, both should work - if no profile is set in the cli argument, the profile in the config should be used. If no default profile exists (only top level config options like in your case, these should be treated as „default“).

The code to treat top level config options as default is probably missing? I trust you to make a call here and proceed with the merging and release :)

libs/foundry-dev-tools/src/foundry_dev_tools/errors/meta.py
```diff
@@ -57,7 +57,9 @@ def __init__(self, response: requests.Response | None = None, info: str | None =
             self.message = api_message
             del self.kwargs["message"]
         term_size = shutil.get_terminal_size().columns
-        msg = f"{self.message}\n{self.info}"
+        msg = self.message
+        if self.info:
+            msg += "\n{self.info}"
```
Silas2701: What is the reason for outsourcing the `self.info` into a separate if-clause? Or was it just something you noticed that had to be fixed anyway?

libs/foundry-dev-tools/src/foundry_dev_tools/cli/config.py
jonas-w: At first I thought it would be a good idea to have profile named "default" as the default config. After I tried various approaches I came to the conclusion that it would need many changes for this to work correctly.

For example all environment variables would need to include "FDT_DEFAULT__" (otherwise I would need to handle this as a special case in the environment variable parser).

Also profiles are intended to inherit from the default config, so it makes more sense to have the default config one "level" above in the toml config. This is useful to set one foundry host name in the default config and then have multiple different credentials for each profile without needing to specify the host all the time. Also this way you don't need to set the config options like `transforms_freeze_cache` for every profile.

> The code to treat top level config options as default is probably missing? I trust you to make a call here and proceed with the merging and release :)

The code right here handles this, if a profile is set it will merge the profile config into the toplevel configuration:
https://github.com/emdgroup/foundry-dev-tools/blob/e50b3d5bee41f3b75b9c40544e454b3f15cc4289/libs/foundry-dev-tools/src/foundry_dev_tools/config/config.py#L151-L162

If no profile is set either as an argument to the FoundryContext or CLI and no 'profile' is set in the toml config it will just return the toplevel config `return config`. This worked, as this line previously returned None:
https://github.com/emdgroup/foundry-dev-tools/blob/e50b3d5bee41f3b75b9c40544e454b3f15cc4289/libs/foundry-dev-tools/src/foundry_dev_tools/config/config.py#L152

libs/foundry-dev-tools/src/foundry_dev_tools/errors/meta.py
jonas-w: Yes I was checking if my error handling also worked for the public API which was the case. But I noticed that every error is prefixed with "None"
image

Silas2701: Ahh alright, I get that

libs/foundry-dev-tools/src/foundry_dev_tools/cli/info.py
```diff
@@ -217,33 +217,41 @@ def _config_section() -> Tree:
...
                 cred_table.add_row(
                     cred_config_entry[1:],
                     _bool_color(
                         value is not None,
-                        f"Is {'not' if value is None else ''} set"
+                        f"Is {'not ' if value is None else ''}set"
                         f"{', but not shown for security reasons' if value is not None else ''}.",
                     ),
                 )
+            elif cred_config_entry.startswith("__"):
```
jonas-w: This check needs to happen before the single underscore check, as a string that starts with double underscores also start with a single underscore 😅

Will fix it

libs/foundry-dev-tools/src/foundry_dev_tools/cli/info.py
```diff
@@ -217,33 +217,41 @@ def _config_section() -> Tree:
...
-        except:  # noqa: S110,E722
-            pass
-        config_node.add(_bool_color(auth, ("Successfully authenticated" if auth else "Failed to authenticate")))
+        username = None
+        with contextlib.suppress(Exception):
+            username = ctx.multipass.api_me().json()["username"]
+        config_node.add(
+            _bool_color(
+                bool(username),
+                (f"Successfully authenticated as '{username}'" if username else "Failed to authenticate"),
```
jonas-w: This "leaks" the username, which would be not that great for public issues opened with the CLI output attached, but I get it that it would be useful to see the authenticated username.

Maybe we could add a CLI option that does not redact the user information like foundry host and username, but the default would be to redact it.

libs/foundry-dev-tools/src/foundry_dev_tools/config/token_provider.py
```diff
@@ -119,7 +119,7 @@ def __init__(
         client_id: str,
         client_secret: str | None = None,
         grant_type: FoundryOAuthGrantType | None = None,
-        scopes: list[str] | None = None,
+        scopes: list[str] | str | None = None,
```
jonas-w: Just so I understand the reason behind this change, when would it be preferable to pass the scopes as a string instead of a list?

EDIT: Oh I see, for environment variables?

libs/foundry-dev-tools/src/foundry_dev_tools/config/token_provider.py
```diff
@@ -142,13 +142,20 @@ def __init__(
         if self.grant_type == "client_credentials" and self._client_secret is None:
             msg = "You need to provide a client secret for the client credentials grant type."
             raise TokenProviderConfigError(msg)
+        scopes = self._scopes_to_list(scopes)
         if self.grant_type == "authorization_code":
             if scopes is not None:
-                self.scopes = list({*DEFAULT_OAUTH_SCOPES, *scopes})
+                self.scopes = scopes
```
jonas-w: In the docstring above we still mention that the scopes get appended, I'll update the docstring/documentation for this.

libs/foundry-dev-tools/src/foundry_dev_tools/config/token_provider.py
nicornk: Yes, I was just testing different options to overwrite scopes and I think I added this so that environment variable passed scopes (comma separated string) get picked up correctly.
I also added a different method to parse true, false as booleans.

libs/foundry-dev-tools/src/foundry_dev_tools/cli/info.py
nicornk: I was testing confits for two stacks and showing the username was helpful, but than I added the foundry url as well to distinguish. I am wondering which is more sensitive… 
One would be needed to understand if the correct config is picked up

libs/foundry-dev-tools/src/foundry_dev_tools/cli/config.py
nicornk: Ok, let’s go with your approach and we can still tweak it later.
I might have done something wrong while testing.

libs/foundry-dev-tools/src/foundry_dev_tools/cli/info.py
jonas-w: Okay I'll leave both in the output, It should be up to the user to decide if username and URL can be public, compared to secrets which should never be public.

PR #53, opened 2024-03-12
Some clarification in documentation for configuration

dasmy:
# Summary

After recognizing that the documentation about configuration with environment variables is not perfectly clear wrt. their capitalization, @nicornk invited me the create a pull request. Here it is. Most notably I tried to go through the whole file and clarify where I had the feeling things were not completely precise or comprehensive.

I tried to verify all changes by inspecting `config.py`. Please double check that I did not misunderstand anything there.

Happy to hear your comments.

# Checklist

- [X] Included tests (or is not applicable).

### DIFF
--- a/docs/Configuration_usage.md
+++ b/docs/Configuration_usage.md
@@ -3,7 +3,7 @@
 Configuration options can be supplied to the library through multiple different methods that are processed in the following order:
   1. Every configuration option is initalized to its default value as given in the [table below](#configuration-options).
   2. If a file `~/.foundry-dev-tools/config` exists, values defined therein overwrite the corresponding default values.
-     See [below](#foundry-dev-toolsconfig-file) for details on its structure.
+     See [below](#foundry-dev-tools-config-file) for details on its structure.
   3. If a project [specific config file](#project-specific-configuration) exists in `your-project-dir/.foundry_dev_tools`, values set in there take precedence.
   4. If an environment variable like `FOUNDRY_DEV_TOOLS_config_key` exists, `config_key` gets overwritten by this environment variable.
      Note, that environment variable names are always all-caps, e.g. the `foundry_url` config value can be set via `export FOUNDRY_DEV_TOOLS_FOUNDRY_URL=https://foundry.example.com`.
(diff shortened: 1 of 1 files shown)


docs/Configuration_usage.md
```diff
@@ -1,32 +1,55 @@
...
+| client_secret                          | The client_secret of the Foundry Third Party application                                                                                          | ...                                    | not set (None)         |
+| grant_type                             | Set this to client_credentials if you use Foundry Third Party application client_credentials flow. Default is authorization_code                  | authorization_code, client_credentials | authorization_code     |
+| scopes                                 | Custom scopes for Third Party Application Tokens                                                                                                  | "api:read-data offline_access"         | not set (None)         |
+| requests_ca_bundle                     | Path to custom ca bundle file, useful in corporate networks with SSL inspection                                                                   | not set                                | `os.path.expanduser()` |
```
jonas-w: `requests_ca_bundle` default value is also unset (None)

docs/Configuration_usage.md
```diff
@@ -1,32 +1,55 @@
...
+| transforms_sql_dataset_size_threshold  | Maximum size of a dataset up to which file gets downloaded. If dataset is larger, sql with limit query will be used to download a partial dataset | 500                                    | 500                    |
+| transforms_sql_sample_select_random    | True or False, if data is randomly sampled when SQL mode is used (query takes more time)                                                          | True or False                          | False                  |
+| transforms_force_full_dataset_download | if file_download is True, all files of datasets are downloaded                                                                                    | True or False                          | False                  |
+| cache_dir                              | Point cache directory to other directory than default                                                                                             | ~/.foundry-dev-tools/cache             | `os.path.expanduser()` |
```
jonas-w: `cache_dir` is per default `~/.foundry-dev-tools/cache`

see here: https://github.com/emdgroup/foundry-dev-tools/blob/5a7d9cd15d9704f4b70f7ff81b6b6b9385aeb1a0/src/foundry_dev_tools/config.py#L117

docs/Configuration_usage.md
```diff
@@ -1,32 +1,55 @@
...
+| transforms_freeze_cache                | Uses offline cache as is; no Foundry connection needed                                                                                            | True or False (default: False)         | False                  |
```
jonas-w: this row shows twice that the value is False, remove it in the `values` column

docs/Configuration_usage.md
```diff
@@ -1,32 +1,55 @@
...
+| transforms_output_folder               | When @transform in combination with `TransformOutput.filesystem()` is used, files are written to this folder.                                     | /projectA/local_files                  | `os.path.expanduser()` |
```
jonas-w: `transforms_output_folder` is also unset (None), per default.

see here: https://github.com/emdgroup/foundry-dev-tools/blob/5a7d9cd15d9704f4b70f7ff81b6b6b9385aeb1a0/src/foundry_dev_tools/config.py#L118

docs/Configuration_usage.md
```diff
@@ -1,32 +1,55 @@
...
+| enable_runtime_token_providers         | Enables the token providers                                                                                                                       | True or False                          | True                   |
+
+## `~/.foundry-dev-tools/config` file
+The content of the file `~/.foundry-dev-tools/config` must be compatible with the Python `configparser` library, i.e. it must have a structure similar to what is found in Microsoft Windows INI files.
```
jonas-w, line 37: Maybe adding a link to what `ConfigParser` really supports could help: https://docs.python.org/3/library/configparser.html#supported-ini-file-structure

docs/Configuration_usage.md
```diff
@@ -1,32 +1,55 @@
...
+[default]
+foundry_url=https://foundry.example.com
+jwt=eyJhb...
+```
+
+## Project specific configuration
+
+You can provide a project-specific configuration via a project-local file `.foundry_dev_tools`.
+This file has to reside in your project's root directory.
+The structure of this file is equivalent to the [`~/.foundry-dev-tools/config`](#foundry-dev-toolsconfig-file) file.
```
jonas-w: link should be: `#foundry-dev-tools-config-file`

docs/Configuration_usage.md
```diff
@@ -1,32 +1,55 @@
 # Configuration
 
-By default, the code loads the configuration from the '~/.foundry-dev-tools/config' file.  
-If a project [specific config file](#project-specific-configuration) exists in `your-project-dir/.foundry_dev_tools`, it takes precedence.  
-If a environment variable like `FOUNDRY_DEV_TOOLS_config_key` exists, `config_key` gets overwritten by this environment variable.  
-Afterwards, the logic falls back to a default value: ( overwrites > environment variable > project config file > config file > default)
+Configuration options can be supplied to the library through multiple different methods that are processed in the following order:
+  1. Every configuration option is initalized to its default value as given in the [table below](#configuration-options).
+  2. If a file `~/.foundry-dev-tools/config` exists, values defined therein overwrite the corresponding default values.
+     See [below](#foundry-dev-toolsconfig-file) for details on its structure.
```
jonas-w: link should be #foundry-dev-tools-config-file

docs/Configuration_usage.md
```diff
@@ -1,32 +1,55 @@
...
+| client_id                              | The client_id of the Foundry Third Party application for SSO Authentication                                                                       | ...                                    | not set (None)         |
+| client_secret                          | The client_secret of the Foundry Third Party application                                                                                          | ...                                    | not set (None)         |
+| grant_type                             | Set this to client_credentials if you use Foundry Third Party application client_credentials flow. Default is authorization_code                  | authorization_code, client_credentials | authorization_code     |
+| scopes                                 | Custom scopes for Third Party Application Tokens                                                                                                  | "api:read-data offline_access"         | not set (None)         |
```
jonas-w: While its true that the `scopes` config option defaults to None, we use these scopes per default if `scopes` is None:

https://github.com/emdgroup/foundry-dev-tools/blob/5a7d9cd15d9704f4b70f7ff81b6b6b9385aeb1a0/src/foundry_dev_tools/foundry_api_client.py#L61-L68

https://github.com/emdgroup/foundry-dev-tools/blob/5a7d9cd15d9704f4b70f7ff81b6b6b9385aeb1a0/src/foundry_dev_tools/foundry_api_client.py#L2311-L2312

But only if `grant_type` is `authorization_code`, as with `client_credentials` we don't pass any scopes to the Foundry API (as we'll just get all scopes this way)

jonas-w (changes requested):
Thanks for the contribution!

I'll make sure to also include this in the v2, see #52

docs/Configuration_usage.md
dasmy: This is indeed interesting: `requests_ca_bundle` is listed in the `TYPES` constant in `config.py` but not in the `config` array where the defaults are set. This means for example that `requests_ca_bundle` cannot be set via an environment variable. Is this intended?

docs/Configuration_usage.md
dasmy: Looking at the code, I am not perfectly sure this will work as expected: Inside config, `scopes` is a string that contains a space-separated list if entries, while `foundry_api_client` seems to assume this to be a list of strings.

I could not find the `scopes.split(' ')` that I would have expected to see anywhere...

Found a `join` for the other direction here, though: https://github.com/emdgroup/foundry-dev-tools/blob/5a7d9cd15d9704f4b70f7ff81b6b6b9385aeb1a0/src/foundry_dev_tools/foundry_api_client.py#L2227

docs/Configuration_usage.md
jonas-w: @dasmy good catch! I don't think this is intended, but otherwise the requests library also has the https://requests.readthedocs.io/en/latest/user/advanced/#ssl-cert-verification `REQUESTS_CA_BUNDLE` environment variable.

docs/Configuration_usage.md
jonas-w: You are right, in theory (according to the types) this shouldn't work, but I dug into `oauthlib` library which the palantir-oauth-client uses, and looked at what happenes with the `scopes` variable.

And it does not get used, until this function gets executed:
https://github.com/oauthlib/oauthlib/blob/f9a07c6c07d0ddac255dd322ef5fc54a7a46366d/oauthlib/oauth2/rfc6749/utils.py#L19

That's why it worked all the time, although this should definitely be changed, as palantir-oauth-client may at some point change behaviour and expect a list of strings.

jonas-w (approved):

PR #56, opened 2024-05-29
Allow .git files for git_toplevel_dir identification to support git submodules

lfleck:
# Summary

For projects with multiple (Foundry code) repositories, it can be handy to add them as [git submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules) to a leading project repository.  In that case, the existing implementation would fail at e.g. the branch identification for transforms in submodules, as the ```git_toplevel_dir``` is assumed to contain a **.git directory**, while submodules only add a **.git file**. 

The proposed solution is to change from ```...(".git").is_dir()``` to ```...(".git").exists()```, allowing files and directories. The respective test case has been extended to cover both scenarios.

# Checklist

### DIFF
--- a/tests/test_utils_repo.py
+++ b/tests/test_utils_repo.py
@@ -23,6 +23,24 @@ def test_git_toplevel_dir(use_git: bool, tmpdir: "py.path.LocalPath"):
         git_dir_cwd = git_toplevel_dir(use_git=use_git)
     assert git_dir_cwd == toplevel
 
+    # Test if git submodules can be recognized as toplevel_dir
+    # Use foundry-dev-tools as dummy submodule
+    subprocess.check_call(
+        [
+            "git",
+            "submodule",
+            "add",
+            "https://github.com/emdgroup/foundry-dev-tools.git",
+            "dummy_submodule",
+        ],
+        cwd=toplevel,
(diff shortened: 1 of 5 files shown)


tests/test_utils_repo.py
```diff
@@ -23,6 +23,24 @@ def test_git_toplevel_dir(use_git: bool, tmpdir: "py.path.LocalPath"):
...
+            "dummy_submodule",
+        ],
+        cwd=toplevel,
+    )
+    toplevel_submodule = toplevel.joinpath("dummy_submodule")
+    git_dir_submodule = git_toplevel_dir(
+        Path(tmpdir.mkdir("dummy_submodule", "submodule_subdirectory")), use_git=use_git
+    )
+    assert git_dir_submodule == toplevel_submodule
+
```
jonas-w: Would it be possible to modify the test so it doesn't require an internet connection? Perhaps by creating a local Git repository to serve as the dummy submodule instead of cloning from GitHub. This would make our tests more robust and self-contained.

Thanks!

lfleck: Good point @jonas-w
I changed it to an empty local repository that gets added as submodule.

PR #63, opened 2024-08-12
Feat/multipass endpoints

Silas2701:
# Summary

Introduce new endpoints to the MultipassClient which have not yet been implemented .

# Checklist

### DIFF
--- a/libs/foundry-dev-tools/src/foundry_dev_tools/clients/multipass.py
+++ b/libs/foundry-dev-tools/src/foundry_dev_tools/clients/multipass.py
@@ -121,7 +121,7 @@ class MultipassClient(APIClient):
 
         return self.api_request("POST", "administration/groups", json=body, **kwargs)
 
-    def api_get_group(self, group_id: str, **kwargs) -> requests.Response:
+    def api_get_group(self, group_id: api_types.GroupId, **kwargs) -> requests.Response:
         """Returns the multipass group information.
 
         Args:
@@ -186,7 +186,7 @@ class MultipassClient(APIClient):
             **kwargs,
         )
 
-    def api_delete_group(self, group_id: str, **kwargs) -> requests.Response:
+    def api_delete_group(self, group_id: api_types.GroupId, **kwargs) -> requests.Response:
         """Deletes multipass group.
 
         Args:
--- a/libs/foundry-dev-tools/src/foundry_dev_tools/clients/multipass.py
+++ b/libs/foundry-dev-tools/src/foundry_dev_tools/clients/multipass.py
@@ -397,6 +397,43 @@ class MultipassClient(APIClient):
             **kwargs,
         )
 
+    def api_get_principals_groups_all(
+        self,
+        principal_ids: set[api_types.PrincipalId],
+        **kwargs,
+    ) -> requests.Response:
+        """Returns Groups that a principal (user or group) is member of.
+
+        Args:
+            principal_ids: The identifiers of the principals to be added to the groups
+            **kwargs: gets passed to :py:meth:`APIClient.api_request`
+
+        Returns:
+            requests.Response:
+                the response contains a json which is a dict in the following shape:
+                   {
+                        "containingGroupIdsByPrincipalId": {
+                            "<requestedPrincipalId>": ["<group1>", ...]
+                        },
+                        "groups": [
+                            {
+                                "id": "<group1>",
+                                "name": "<group1Name>"
+                            }
+                        ]
+                   }
+        """
+        body = {
+            "principalIds": list(principal_ids),
+        }
+
+        return self.api_request(
+            "POST",
+            "administration/principals/bulk/groups/all",
+            json=body,
+            **kwargs,
+        )
+
     def api_get_immediate_group_members(
         self,
         group_id: api_types.GroupId,
--- a/libs/foundry-dev-tools/src/foundry_dev_tools/clients/multipass.py
+++ b/libs/foundry-dev-tools/src/foundry_dev_tools/clients/multipass.py
@@ -57,11 +57,11 @@ class MultipassClient(APIClient):
 
     api_name = "multipass"
 
-    def get_user_info(self) -> dict:
+    def get_user_info(self) -> api_types.User:
         """Returns the json dict from the :py:meth:`foundry_dev_tools.clients.multipass.MultipassClient.api_me` API.
 
         Returns:
-            dict:
+            api_types.User:
 
         .. code-block:: python
 
@@ -299,7 +299,7 @@ class MultipassClient(APIClient):
     def add_group_manager_managers(
         self,
         group_id: api_types.GroupId,
-        manager_managers: set[api_types.PrincipalId] | None = None,
+        manager_managers: set[api_types.PrincipalId],
     ) -> requests.Response:
         """Assign principals as manager managers for the specified group.
 
@@ -312,7 +312,7 @@ class MultipassClient(APIClient):
     def add_group_member_managers(
         self,
         group_id: api_types.GroupId,
-        member_managers: set[api_types.PrincipalId] | None = None,
+        member_managers: set[api_types.PrincipalId],
     ) -> requests.Response:
         """Assign principals as member managers for the specified group.
 
@@ -325,7 +325,7 @@ class MultipassClient(APIClient):
     def remove_group_manager_managers(
         self,
         group_id: api_types.GroupId,
-        manager_managers: set[api_types.PrincipalId] | None = None,
+        manager_managers: set[api_types.PrincipalId],
     ) -> requests.Response:
         """Withdraw the provided principals from the manager managers for the specified group.
 
@@ -338,7 +338,7 @@ class MultipassClient(APIClient):
     def remove_group_member_managers(
         self,
         group_id: api_types.GroupId,
-        member_managers: set[api_types.PrincipalId] | None = None,
+        member_managers: set[api_types.PrincipalId],
     ) -> requests.Response:
         """Withdraw the provided principals from the member managers for the specified group.
 
@@ -597,15 +597,15 @@ class MultipassClient(APIClient):
     def reset_group_member_expiration_settings(
         self,
         group_id: api_types.GroupId,
-    ) -> bool:
+    ) -> api_types.GroupMemberExpirationSettings:
         """Reset the group member expiration settings for the specified group.
 
         Args:
             group_id: The identifier of the group for which to restore the initial state of the expiration settings
 
         Returns:
-            bool:
-                indicator whether the group member expiration settings have successfully been reset
+            api_types.GroupMemberExpirationSettings:
+                the updated group member expiration settings
         """
         resp = self.api_update_group_member_expiration_settings(
             group_id, max_expiration=None, max_duration_in_seconds=None
--- a/libs/foundry-dev-tools/src/foundry_dev_tools/clients/multipass.py
+++ b/libs/foundry-dev-tools/src/foundry_dev_tools/clients/multipass.py
@@ -1031,6 +1031,33 @@ class MultipassClient(APIClient):
             **kwargs,
         )
 
+    def api_get_groups(self, **kwargs) -> requests.Response:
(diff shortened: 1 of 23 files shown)


libs/foundry-dev-tools/src/foundry_dev_tools/clients/multipass.py
```diff
@@ -13,25 +14,54 @@
...
+
+    Returns:
+        datetime:
+            the datetime object with the time zone information
+    """
+    if not dt.tzinfo:
+        msg = (
+            f"{dt.isoformat(timespec='seconds')} passed for parameter '{variable_name} is missing timezone information."
+            f"Ensure timezone is provided for datetime object. Defaults to '{timezone.utc!s}' timezone!"
+        )
```
jonas-w:
```suggestion
        msg = (
            f"{dt.isoformat(timespec='seconds')} passed for parameter '{variable_name} is missing timezone information."
            f"Setting the timezone to '{timezone.utc!s}'."
            f"Ensure timezone is provided for datetime object."
        )
```

Reworded it to hopefully make it more clear that the code automatically sets the timezone to UTC

libs/foundry-dev-tools/src/foundry_dev_tools/clients/multipass.py
```diff
@@ -105,6 +217,402 @@ def api_delete_group(self, group_id: str, **kwargs) -> requests.Response:
...
+                    msg = (
+                        f"expiration value '{validated_expiration.isoformat(timespec='seconds')}' "
+                        f"for 'expirations['{group_id}']['{principal_id}']' must not be in past!"
+                    )
+                    raise ValueError(msg)
+
+                body["expirations"].setdefault(group_id, {})[principal_id] = validated_expiration.isoformat(
+                    timespec="seconds"
+                )
+
```
jonas-w:
```suggestion
        if expirations:
            for group_id, expiration_mapping in expirations.items():
                group_expirations = {}

                for principal_id, expiration in expiration_mapping.items():
                    validated_expiration = _validate_timezone(
                        expiration, f"expirations['{group_id}']['{principal_id}']"
                    )

                    if validated_expiration <= datetime.now(validated_expiration.tzinfo):
                        msg = (
                            f"expiration value '{validated_expiration.isoformat(timespec='seconds')}' "
                            f"for 'expirations['{group_id}']['{principal_id}']' must not be in past!"
                        )
                        raise ValueError(msg)

                    group_expirations[principal_id] = validated_expiration.isoformat(timespec="seconds")

                body["expirations"][group_id] = group_expirations
```

Cleaned this code up to first check if it is needed to run the for loop, and then removed the setdefault and replaced it with one dict per outer for loop iteration which then gets set in the expirations dictionary

libs/foundry-dev-tools/src/foundry_dev_tools/clients/multipass.py
```diff
@@ -105,6 +217,402 @@ def api_delete_group(self, group_id: str, **kwargs) -> requests.Response:
...
+            max_duration_in_seconds: When adding a new membership, it can last no longer than
+                the specified maximum duration. Expiration of existing memberships will be adjusted accordingly.
+                Value passed must be greater equal :py:const:`MINIMUM_MAX_DURATION_IN_SECONDS` and defaults to
+                :py:const:`DEFAULT_MAX_DURATION_IN_SECONDS` if it does not meet the condition
+            **kwargs: gets passed to :py:meth:`APIClient.api_request`
+
+        Returns:
+            requests.Response:
+                the response contains a json which holds the updated expiration settings
+        """
```
jonas-w: Explain what happens when max_expiration/max_duration_in_seconds is set to None, and make it clear that if max_duration_in_seconds is set to None that it won't get set to the default_max_duration_in_seconds

libs/foundry-dev-tools/src/foundry_dev_tools/clients/multipass.py
```diff
@@ -105,6 +217,402 @@ def api_delete_group(self, group_id: str, **kwargs) -> requests.Response:
...
+        if max_expiration:
+            max_expiration = _validate_timezone(max_expiration, "max_expiration")
+            now = datetime.now(max_expiration.tzinfo)
+
+            if max_expiration <= now:
+                new_expiration = (now + timedelta(days=1)).replace(hour=0, minute=0, second=0, microsecond=0)
```
jonas-w: Is there a specific reason why this gets set to now+1day, and why it does not simply fail? (Could be 'problematic' if the user does somehow not see the warning and thinks that the expiration was set to the time specified).

If this behaviour should be kept, then please explain it in the docstring that this can happen.

libs/foundry-dev-tools/src/foundry_dev_tools/clients/multipass.py
```diff
@@ -13,25 +14,54 @@
...
+            f"Ensure timezone is provided for datetime object. Defaults to '{timezone.utc!s}' timezone!"
+        )
+        warnings.warn(msg)
+
+        dt = dt.replace(tzinfo=timezone.utc)
+
+    return dt
+
 
 # PLACEHOLDER
```
jonas-w: remove this one

libs/foundry-dev-tools/src/foundry_dev_tools/clients/multipass.py
```diff
@@ -105,6 +217,402 @@ def api_delete_group(self, group_id: str, **kwargs) -> requests.Response:
...
+        Args:
+            group_ids: A set of group identifiers principals should be added to
+            principal_ids: The identifiers of the principals to be added to the groups
+            expirations: Optional expiration settings that can be passed
+                if principals should only have temporal access to groups
+            **kwargs: gets passed to :py:meth:`APIClient.api_request`
+        """
+        body = {
+            "groupIds": list(group_ids) if group_ids else None,
+            "principalIds": list(principal_ids) if principal_ids else None,
```
jonas-w: What does groupIds/principalIds set to None do? This seems like undefined behaviour?

As group_ids/principal_ids can't be None according to the function definition, this could also be replaced with just list(group_ids)/list(principal_ids), which would then yield an empty array, but nonetheless what would happen if either one of this is None/empty?

libs/foundry-dev-tools/src/foundry_dev_tools/resources/__init__.py
```diff
@@ -1,5 +1,8 @@
 from foundry_dev_tools.resources.dataset import Dataset
 from foundry_dev_tools.resources.folder import Folder
+from foundry_dev_tools.resources.group import Group
+from foundry_dev_tools.resources.principal import Principal
 from foundry_dev_tools.resources.resource import Resource
+from foundry_dev_tools.resources.user import User
 
-__all__ = ["Resource", "Dataset", "Folder"]
+__all__ = ["Resource", "Dataset", "Folder", "Group", "Principal", "User"]
```
jonas-w: The foundry_dev_tools.resources module is only intended for resources, as Group/User does not have an rid they are not 'resources'.

Maybe we could create another module for these 'helper classes' which are not resources?

libs/foundry-dev-tools/src/foundry_dev_tools/utils/api_types.py
```diff
@@ -477,3 +489,26 @@ class ProjectFolderDisplaySettingsUpdate(TypedDict):
...
+    """Foundry multipass group."""
+
+    id: GroupId
+    name: str
+    attributes: dict[str, list[str]]
+
+
+class GroupMemberExpirationSettings(TypedDict):
+    """Foundry GroupMemberExpirationSettings API object."""
+
```
jonas-w, line 515: I personally find these variables a bit confusing, so adding a docstring to this definiton here, too (not only in the function which uses this) could be beneficial?

Or is there even some public Foundry Documentation we could link to?

tests/integration/utils.py
```diff
@@ -285,8 +292,65 @@
...
+
+
+def backoff(
+    func: Callable[[], T],
+    predicate: Callable[[T], bool],
+    sleep: float = INITIAL_SLEEP_TIME,
+    num_retries: int = MAX_RETRIES,
+) -> tuple[bool, T, int]:
+    result = func()
+
```
jonas-w, line 338: Could you add a docstring for this method for future reference? (So when someone looks at this code in the future it is more immediately clear how to use it/what it does)

jonas-w (changes requested):

tests/integration/clients/test_multipass.py
```diff
@@ -3,28 +3,182 @@
...
 
+from datetime import datetime, timedelta, timezone
+from random import choice
+from string import ascii_uppercase
+
 import pytest
 import requests
-
-from foundry_dev_tools.clients.multipass import DEFAULT_TOKEN_LIFETIME_IN_SECONDS
+from integration.utils import (
```
jonas-w: the module should be `tests.integration.utils`, `integration.utils` throws an error that it does not exist when run via pytest.

tests/integration/resources/test_group.py
```diff
@@ -0,0 +1,68 @@
+from random import choice
+from string import ascii_uppercase
+
+import requests
+from integration.conftest import TEST_SINGLETON
+from integration.utils import skip_test_on_error
```
jonas-w: Same issue here, should be `tests.integration....`

tests/integration/resources/test_principal.py
```diff
@@ -0,0 +1,23 @@
+import requests
+from integration.conftest import TEST_SINGLETON
+from integration.utils import skip_test_on_error
```
jonas-w: Same issue here, should be tests.integration....

jonas-w (changes requested):

libs/foundry-dev-tools/src/foundry_dev_tools/clients/multipass.py
Silas2701: It is the default behaviour of Foundry setting the expiration date to the beginning of the next day (00:00:00) when the value is invalid. But you are right, failing would even be a better option!

libs/foundry-dev-tools/src/foundry_dev_tools/clients/multipass.py
Silas2701: That's true, it does not make sense and is probably just a mistake which I oversaw when checking again the functions I implemented. Thanks for the hint!

If the list of group ids was None or empty, the principals wouldn't be added as members to the list of groups. If the list of principals was empty, also no members would be added to groups, so it isn't undefined behaviour.
@jonas-w But do you think a Non-empty check should be included for group ids and principal ids to assert that those sets are not empty? (Idk if it is necessary to have, as other API calls also do not check for empty sets)

libs/foundry-dev-tools/src/foundry_dev_tools/clients/multipass.py
Silas2701: Great, looks neat and clean! I was already looking for a solution like that, as I felt a little uncomfortable with my previous approach

libs/foundry-dev-tools/src/foundry_dev_tools/resources/__init__.py
Silas2701: True, I didn't think about the fact that Group/User are actually not resources until now. But it would be good to move them to another module for which I don not know a name yet. If you have suggestions let me know!

libs/foundry-dev-tools/src/foundry_dev_tools/utils/api_types.py
Silas2701: There is even some public Foundry Documentation which we could link to!

libs/foundry-dev-tools/src/foundry_dev_tools/clients/multipass.py
Silas2701: Addition: I just tested the behaviour for empty group ids and/or principals and I don't think it is necessary to have a non-empty check. It is simply just like as if it wouldn't execute anything.

PR #106, opened 2025-07-14
User Friendly error message when jwt token is invalid or expired

### LINKED ISSUE #95: [Feature Request]: More user-friendly error message when token is invalid/expired
### Issue checklist

- [x] This is a **feature request**/**enhancement**. And not a bug.
- [x] I searched through the GitHub issues and this feature/enhancement has not been requested before.
- [x] I have installed the latest version of Foundry DevTools and don't use an unsupported python version.
- [x] Others could also benefit from this feature or enhancement and it is **not** a very specific use case.

### Feature use case

The error message that people receive when they have configured an invalid or expired token is not self-explanatory. Several people already asked me in the past why they don't have permission for a specific dataset anymore when in fact they had permission but an expired token, but they got misled by the error message. It would be great if foundry-dev-tools would return a custom error message such as `"Could not authenticate <your-email> at <your-configured-host>. Please check if your jwt token is invalid or has expired."` 

### Description of the Feature

This is the error message that people currently receive with an invalid token:
```
FoundryAPIError: Details about the Foundry API error:

------------------------------------REQUEST------------------------------------
METHOD = POST
CONTENT-TYPE = application/json
ENDPOINT = /foundry-sql-server/api/queries/execute
-----------------------------------PARAMETERS-----------------------------------
branch = master
dialect = SPARK
timeout = 600
error_code = PERMISSION_DENIED
error_name = FoundrySqlServer:NotAuthorized
error_instance_id = b1b17d38-3b9f-4d8f-9089-xxx
operations = []
------------------------------------RESPONSE------------------------------------
ERROR_NAME = FoundrySqlServer:NotAuthorized
ERROR_CODE = PERMISSION_DENIED
ERROR_INSTANCE_ID = b1b17d38-3b9f-4d8f-9089-xxx
STATUS = 403
```

The error message can be reproduced with the code example from the docs:
```python
from foundry_dev_tools import FoundryContext

ctx = FoundryContext()
df = ctx.foundry_sql_server.query_foundry_sql(
    "SELECT * FROM `/Global/Foundry Training and Resources/Example Data/Aviation Ontology/airlines`",
    branch="master",
)
```

As mentioned, it would be great if people would instead receive an easier to interpret message.

### Alternatives you considered

I use `fdt info` to check if authentication works. But I can imagine that not everybody uses the CLI tool. In addition, if people do not even get the idea that their issue is authentication related, they would not verify this in the first place.

### Additional Context

_No response_

jonas-w:
# Summary

This should fix #95

- JWTTokenProvider now has an `__expiration` attribute.
- this attribute gets set by the `set_expiration` method, which does the following:
1. parses the claims from the jwt -> if this fails, `InvalidOrExpiredJWTTokenError`
2. converts the `jti` claim to an UUID
3. retrieves all user generated tokens via `get_tokens` -> if this API call fails, the exception will be thrown too, as this means the JWT Token is expired
4. searches for the token_id in the retrieved tokens, if matched, then the `__expiration` attribute will be set

- the `token()` property, now checks for the `__expiration` time if it's set

# Checklist

### DIFF
--- a/libs/foundry-dev-tools/src/foundry_dev_tools/config/token_provider.py
+++ b/libs/foundry-dev-tools/src/foundry_dev_tools/config/token_provider.py
@@ -3,8 +3,9 @@
 from __future__ import annotations
 
 import base64
+import json
 import time
-from functools import cached_property
+import uuid
 from typing import TYPE_CHECKING, ClassVar
 
 import palantir_oauth_client
@@ -12,13 +13,14 @@ import requests
 from requests.structures import CaseInsensitiveDict
 
 from foundry_dev_tools.config.config_types import Host
-from foundry_dev_tools.errors.config import TokenProviderConfigError
+from foundry_dev_tools.errors.config import InvalidOrExpiredJWTTokenError, TokenProviderConfigError
 from foundry_dev_tools.errors.handling import ErrorHandlingConfig, raise_foundry_api_error
 from foundry_dev_tools.errors.multipass import ClientAuthenticationFailedError
 from foundry_dev_tools.utils.config import entry_point_fdt_token_provider
 
 if TYPE_CHECKING:
     from foundry_dev_tools.config.config_types import FoundryOAuthGrantType, Token
+    from foundry_dev_tools.config.context import FoundryContext
 
 
 class TokenProvider:
@@ -56,9 +58,17 @@ class TokenProvider:
     def set_requests_session(self, session: requests.Session) -> None:
         """No-op by default."""
 
+    def set_expiration(self, ctx: FoundryContext) -> None:
+        """No-op by default.
+
+        Used for jwt expiration retrieval.
+        """
+
 
 class JWTTokenProvider(TokenProvider):
     """Provides Host and Token."""
+
+    __expiration: int | None = None
 
     def __init__(self, host: Host | str, jwt: Token) -> None:
         """Initialize the JWTTokenProvider.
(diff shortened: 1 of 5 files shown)


libs/foundry-dev-tools/src/foundry_dev_tools/config/token_provider.py
```diff
@@ -70,9 +80,39 @@ def __init__(self, host: Host | str, jwt: Token) -> None:
...
+        except Exception as e:
+            raise InvalidOrExpiredJWTTokenError from e
+
+    def set_expiration(self, ctx: FoundryContext) -> None:
+        """This method decodes the jwt token and retrieves its expiration."""
+        if self.__expiration:
+            return
+        token_id = self._decode_token_id()
+        try:
+            tokens = ctx.multipass.get_tokens()  # retrieve all tokens the user has generated
```
nicornk, line 102: What's the performance penalty of this call?
Wouldn't this call fail in case the token used to make this request is already expired?

jonas-w: Yes if this call fails, as mentioned in the PR, the same exception will be thrown (this call is inside a try except block, which on except throws the InvalidOrExpiredJWTTokenError).

This will only be called on creation of the FoundryContext, so it will only increase startup time and not decrease performance for every api request, and for me it takes around 300-500ms for this init step (which is the same time it takes for every other api call with my internet connection)

jonas-w: this could theoretically be cached, so per jwt it would only need to do this request one time, and then save it into a json file, which holds all expiration times for each token id

PR #111, opened 2025-10-24
Feat/public v2 api datasets/nm

nmohr192:
# Summary

Extend the `FoundryContext` with a cached client `public_sdk` for using the public V2 APIs for dataset interaction. This enables foundry dev tools to read datasets even with user access tokens coming from OSDK enabled TPAs.

# Checklist

### DIFF
--- a/libs/foundry-dev-tools/src/foundry_dev_tools/__init__.py
+++ b/libs/foundry-dev-tools/src/foundry_dev_tools/__init__.py
@@ -7,6 +7,7 @@ from foundry_dev_tools.config.config_types import Host
 from foundry_dev_tools.config.context import FoundryContext
 from foundry_dev_tools.config.token_provider import JWTTokenProvider, OAuthTokenProvider
 from foundry_dev_tools.foundry_api_client import FoundryRestClient
+from foundry_dev_tools.public_sdk import PublicSDK
 
 __all__ = [
     "__version__",
@@ -14,6 +15,7 @@ __all__ = [
     "Host",
     "JWTTokenProvider",
     "FoundryContext",
+    "PublicSDK",
     "OAuthTokenProvider",
     "FoundryRestClient",
     "CachedFoundryClient",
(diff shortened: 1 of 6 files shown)


nicornk:
I was thinking about not implementing all functions of the public SDK again, but rather provide a method on the FoundryContext that returns an instance of foundry_sdk.v2.FoundryClient

```


@cached_property
def public_client_v2(self) -> FoundryClient
    """Returns :py:class:`foundry_sdk.v2.FoundryClient`."""
     from foundry_sdk.v2 import FoundryClient
    return FoundryClient(
    auth=FoundryDevToolsAuth(self)
    hostname=self.host
)
```

where we implement FoundryDevToolsAuth based on the Auth interface that is defined here:
https://github.com/palantir/foundry-platform-python/blob/develop/foundry_sdk/_core/auth_utils.py#L38
FoundryDevToolsAuth could hold the Context and return get_token from the current token method.

```
from foundry_sdkimport Auth
# Didn't find a public import for below, maybe we just need to copy it, or remove it as parent class
from foundry_sdk._core.auth_utils import Token

class FoundryDevToolsToken(Token):

    def __init__(self, token: str) -> None:
        self._token = token

    @property
    def access_token(self) -> str:
        return self._token

class FoundryDevToolsAuth(Auth):
    def __init__(self, ctx: FoundryContext) -> None:
        self._ctx = ctx

    def get_token(self) -> Token:
        return FoundryDevToolsToken(self._ctx.token)

    def execute_with_token(self, func: Callable[[Token], T]) -> T:
        return func(self.get_token())

    def run_with_token(self, func: Callable[[Token], T]) -> None:
        func(self.get_token())
```

That way users could just use all methods of the official sdk but still maintain the integrations with our companies infra.

nmohr192:
@nicornk I stumbled upon the public sdk only earlier and actually wanted to talk to you about it. I'll change my PR.

would you add the official sdk as an optional dependency group?
and part of `full`? 

https://pypi.org/project/foundry-platform-sdk/

```toml
[project.optional-dependencies]
s3 = ["aiobotocore[boto3]"]
full = ["foundry-dev-tools-transforms", "foundry-dev-tools[s3]"]
```

nicornk:
> @nicornk I stumbled upon the public sdk only earlier and actually wanted to talk to you about it. I'll change my PR.
> 
> would you add the official sdk as an optional dependency group? and part of `full`?
> 
> https://pypi.org/project/foundry-platform-sdk/
> 
> ```toml
> [project.optional-dependencies]
> s3 = ["aiobotocore[boto3]"]
> full = ["foundry-dev-tools-transforms", "foundry-dev-tools[s3]"]
> ```

Agree on your dependency proposal.

nmohr192:
closed in favor of #112

PR #112, opened 2025-10-27
Integrate public foundry sdk with fdt authentication

nmohr192:
# Summary

Extend the FoundryContext with a cached client `public_client_v2` for using the public V2 APIs for dataset interaction. This enables foundry dev tools to read datasets even with user access tokens coming from OSDK enabled TPAs.

# Checklist

### DIFF
--- a/libs/foundry-dev-tools/src/foundry_dev_tools/config/context.py
+++ b/libs/foundry-dev-tools/src/foundry_dev_tools/config/context.py
@@ -198,7 +198,7 @@ class FoundryContext:
 
         return FoundryClient(
             auth=FoundryDevToolsAuth(self),
-            hostname=self.host,
+            hostname=self.host.domain,
         )
 
     def get_dataset(
--- a/libs/foundry-dev-tools/src/foundry_dev_tools/config/context.py
+++ b/libs/foundry-dev-tools/src/foundry_dev_tools/config/context.py
@@ -196,6 +196,15 @@ class FoundryContext:
         """Returns :py:class:`foundry_sdk.v2.FoundryClient`.
 
         Uses fdt context to integrate with the official foundry-platform-sdk.
+
+        Examples:
+            >>> from foundry_dev_tools._optional.polars import pl
+            >>> from foundry_dev_tools._optional.pyarrow import pa
+
+            >>> pub_client = ctx.public_client_v2
+            >>> ds = pub_client.datasets.Dataset.read_table(rid, format="ARROW")
+            >>> pa_df = pa.ipc.open_stream(ds).read_all()
+            >>> polars_df = pl.from_arrow(pa_df)
         """
         from foundry_dev_tools.public_sdk.auth import FoundryDevToolsAuth
 
--- a/libs/foundry-dev-tools/src/foundry_dev_tools/config/context.py
+++ b/libs/foundry-dev-tools/src/foundry_dev_tools/config/context.py
@@ -33,18 +33,13 @@ from foundry_dev_tools.resources.resource import Resource
 from foundry_dev_tools.utils.config import entry_point_fdt_api_client
 
 if TYPE_CHECKING:
-    import foundry_sdk.v2
-    from foundry_sdk.v2 import FoundryClient
+    from foundry_sdk.v2 import FoundryClient as FoundrySdkV2Client
 
     from foundry_dev_tools.cached_foundry_client import CachedFoundryClient
     from foundry_dev_tools.config.config_types import Host, Token
     from foundry_dev_tools.config.token_provider import TokenProvider
     from foundry_dev_tools.foundry_api_client import FoundryRestClient
     from foundry_dev_tools.utils import api_types
-else:
-    from foundry_dev_tools._optional import foundry_platform_sdk
-
-    foundry_sdk = foundry_platform_sdk.foundry_sdk
 
 
 class FoundryContext:
@@ -192,7 +187,7 @@ class FoundryContext:
         return FoundryRestClient(ctx=self)
 
     @cached_property
-    def public_client_v2(self) -> FoundryClient:
+    def public_client_v2(self) -> FoundrySdkV2Client:
         """Returns :py:class:`foundry_sdk.v2.FoundryClient`.
 
         Uses fdt context to integrate with the official foundry-platform-sdk.
@@ -206,6 +201,7 @@ class FoundryContext:
             >>> pa_df = pa.ipc.open_stream(ds).read_all()
             >>> polars_df = pl.from_arrow(pa_df)
         """
+        from foundry_dev_tools._optional.foundry_platform_sdk import foundry_sdk
         from foundry_dev_tools.public_sdk.auth import FoundryDevToolsAuth
 
         return foundry_sdk.v2.FoundryClient(
--- a/libs/foundry-dev-tools/src/foundry_dev_tools/config/context.py
+++ b/libs/foundry-dev-tools/src/foundry_dev_tools/config/context.py
@@ -193,8 +193,8 @@ class FoundryContext:
         Uses fdt context to integrate with the official foundry-platform-sdk.
 
         Examples:
-            >>> from foundry_dev_tools._optional.polars import pl
-            >>> from foundry_dev_tools._optional.pyarrow import pa
+            >>> import polars as pl
+            >>> import pyarrow as pa
 
             >>> pub_client = ctx.public_client_v2
             >>> ds = pub_client.datasets.Dataset.read_table(rid, format="ARROW")
--- a/libs/foundry-dev-tools/src/foundry_dev_tools/config/context.py
+++ b/libs/foundry-dev-tools/src/foundry_dev_tools/config/context.py
@@ -33,6 +33,8 @@ from foundry_dev_tools.resources.resource import Resource
 from foundry_dev_tools.utils.config import entry_point_fdt_api_client
 
 if TYPE_CHECKING:
+    from foundry_sdk.v2 import FoundryClient
+
     from foundry_dev_tools.cached_foundry_client import CachedFoundryClient
     from foundry_dev_tools.config.config_types import Host, Token
     from foundry_dev_tools.config.token_provider import TokenProvider
@@ -183,6 +185,21 @@ class FoundryContext:
         from foundry_dev_tools.foundry_api_client import FoundryRestClient
 
         return FoundryRestClient(ctx=self)
+
+    @cached_property
+    def public_client_v2(self) -> FoundryClient:
+        """Returns :py:class:`foundry_sdk.v2.FoundryClient`.
+
+        Uses fdt context to integrate with the official foundry-platform-sdk.
+        """
+        from foundry_sdk.v2 import FoundryClient
+
+        from foundry_dev_tools.public_sdk.auth import FoundryDevToolsAuth
+
+        return FoundryClient(
+            auth=FoundryDevToolsAuth(self),
+            hostname=self.host,
+        )
 
     def get_dataset(
         self,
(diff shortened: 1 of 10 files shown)


nmohr192:
@nicornk please review. Once done, I will update the changelog and create the changelog commit.

I know this is the cleanest separation between fdt and the foundry sdk, but I was wondering whether it makes sense to subclass `from foundry_sdk.v2.FoundryClient` and add helper methods for e.g. dataset interaction. Currently, to read a dataset using the public SDK, you need the following code. I assume reading tabular datasets is one of the top use-cases of fdt.

```python
from foundry_dev_tools._optional.polars import pl
from foundry_dev_tools._optional.pyarrow import pa

pub_client = ctx.public_client_v2
ds = pub_client.datasets.Dataset.read_table(rid, format="ARROW")
pa_df = pa.ipc.open_stream(ds).read_all()
polars_df = pl.from_arrow(pa_df)
```


Alternatively, I would at least add the example to the docstring of the `FoundryContext.public_client_v2` property:

```python
    @cached_property
    def public_client_v2(self) -> FoundryClient:
        """Returns :py:class:`foundry_sdk.v2.FoundryClient`.

        Uses fdt context to integrate with the official foundry-platform-sdk.
        """
        from foundry_sdk.v2 import FoundryClient

        from foundry_dev_tools.public_sdk.auth import FoundryDevToolsAuth

        return FoundryClient(
            auth=FoundryDevToolsAuth(self),
            hostname=self.host.domain,
        )
```

jonas-w:
@nmohr192 Would you mind using the existing _optional infrastructure for the foundry_sdk dependency, as we are already using it for other optional dependencies, this way there is a more user friendly message when trying to use this feature without the sdk installed, explaining that it's an optional dependency.

https://github.com/emdgroup/foundry-dev-tools/tree/b09aaaa137e4e3c3a3eb76b5e12fd85d2ccd52e3/libs/foundry-dev-tools/src/foundry_dev_tools/_optional

libs/foundry-dev-tools/src/foundry_dev_tools/public_sdk/auth.py
```diff
@@ -0,0 +1,76 @@
+"""Auth adapters for integrating FoundryDevTools with foundry-platform-sdk."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, TypeVar
+
+from foundry_sdk import Auth
+from foundry_sdk._core.auth_utils import Token
```
nicornk: @nmohr192 can we avoid depending on a private module by copying over some code?

nicornk:
> @nicornk please review. Once done, I will update the changelog and create the changelog commit.
> 
> I know this is the cleanest separation between fdt and the foundry sdk, but I was wondering whether it makes sense to subclass `from foundry_sdk.v2.FoundryClient` and add helper methods for e.g. dataset interaction. Currently, to read a dataset using the public SDK, you need the following code. I assume reading tabular datasets is one of the top use-cases of fdt.

Please add the example to the docstring, good idea.

Regarding reading datasets, I could think off multiple ways to achieve it. We could have a global toggle on the FoundryContext `prefer_public_apis` (or the likes) and than in the normal dataset access methods or e.g. foundry_sql_server calls we use the public SDK code. An alterntive is to add parameters to each of those functions `use_public_api`.

Maybe a combination of both.

@jonas-w  Thoughts from your side?

libs/foundry-dev-tools/src/foundry_dev_tools/public_sdk/auth.py
```diff
@@ -0,0 +1,76 @@
...
+
+        Args:
+            func: Function to execute with the token
+
+        Returns:
+            T: The return value of the function
+        """
+        return func(self.get_token())
+
+    def run_with_token(self, func: Callable[[FoundryDevToolsToken], T]) -> None:
```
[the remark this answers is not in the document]
nmohr192: This is taken 1:1 from the abstract class defined in the palantir SDK:
https://github.com/palantir/foundry-platform-python/blob/develop/foundry_sdk/_core/auth_utils.py#L38

jonas-w:
I think these resources classes could be a good start to integrate the public APIs:
https://github.com/emdgroup/foundry-dev-tools/tree/main/libs/foundry-dev-tools/src/foundry_dev_tools/resources



> Regarding reading datasets, I could think off multiple ways to achieve it. We could have a global toggle on the FoundryContext `prefer_public_apis` (or the likes) and than in the normal dataset access methods or e.g. foundry_sql_server calls we use the public SDK code. An alterntive is to add parameters to each of those functions `use_public_api`.

What would be the use case to implement a toggle for the public APIs? I'm open for that, but implementing the code for a method twice (public + private) seems redundant to me.

nicornk:
The public APIs require other scopes, I am not sure we don’t break users code if we just change it.
But worth giving it a try and taking the risk.

libs/foundry-dev-tools/src/foundry_dev_tools/public_sdk/auth.py
nicornk: @nmohr192 can you resolve this so we can merge?

nicornk (changes requested):

libs/foundry-dev-tools/src/foundry_dev_tools/public_sdk/auth.py
nicornk: Just realized I didn’t complete the review. Sorry

nmohr192: I am looking into this right now... maybe a dumb question but how do you cleanly build this?
I understand how to do it, but it does not integrate nicely with the code/IDE.

Look at this example:
image

You lose all typehinting in the IDE because the IDE can't understand whether pyspark is now a FakeModule or the real module.

So if I build this fakemodule wrapper here for the foundry_sdk:
image

Downstream code like within auth.py becomes ugly, as I can't directly import `Token` anymore for example:
`from foundry_dev_tools._optional.foundry_platform_sdk.foundry_sdk._core.auth_utils import Token` doesn't work.
image

Within this helper class `auth.py` I could probably also import from the real foundry_sdk module instead of the fakemodule wrapper, as it would be imported after the fakemodule one within the `context.py`. But this is also inconsistent not pretty.

image

nmohr192: One solution to have the module initialization done with the FakeModule wrapper + still keeping IDE support would be conditional imports, even though this is also a bit ugly:

image
image


---


Same here:
image
image


Opinions?

nmohr192: a third option would be not having only a FakeModule wrapper around the root level `foundry_sdk` module, but one for each of the downstream modules we use (Auth, Token, FoundryClient). But this would break your pattern of one script + one fake module per library (pandas, polars, pyspark, etc.)

nmohr192:
> The public APIs require other scopes, I am not sure we don’t break users code if we just change it. But worth giving it a try and taking the risk.

Are you saying that we should test the dataset -> polars functionality with an appservice TPA usertoken, and if it also works, we completely replace the `Datasets` class in the [resources](https://github.com/emdgroup/foundry-dev-tools/blob/main/libs/foundry-dev-tools/src/foundry_dev_tools/resources/dataset.py) module with public API calls under the hood? 

This needs to be a followup PR then, I am not sure if everything can be replicated with the public APIs.

libs/foundry-dev-tools/src/foundry_dev_tools/public_sdk/auth.py
jonas-w: @nmohr192 what IDE are you using? As I use pylsp with helix, and I can use autocompletion, so I wasn't aware of this problem when creating this solution.

nmohr192: @jonas-w vscode with pylance

jonas-w: @nmohr192 I just installed vscode and pylance, and it works for me there, too. Are you sure that you have the optional dependency in your environment and vscode points to that environment?

image

EDIT: nevermind I just realised my issue, when I use `from pyspark...` it probably uses the normal pyspark module.

The first level e.g. pyspark.sql autocompletes, but after pyspark.sql it stops autocompleting, which works in my pylsp+helix setup, but not in vscode+pylance apparently.

This seems to be a general issue with pylance, e.g. if I have a `foo.py` with
```
import pyspark
```

and in a `bar.py` I import it from there:
```
from foo import pyspark
```
It will only autocomplete pyspark.X but not pyspark.X.Y.

So the `FakeModule` doesn't seem to be the problem, but rather importing a module from another file where its already imported?

nicornk: `from foundry_sdk._core.auth_utils import Token`

Alternative we can live with this private import it for now and open a PR on https://github.com/palantir/foundry-platform-python/blob/develop/foundry_sdk/_core/auth_utils.py#L27 and ask for Token to be added to the public API.

nmohr192: @jonas-w yeah, that was exactly my struggle as well. I sticked to your implementation now, and ensured (with the test cases, testing with the -public optional dependency group and without it), that the behavior is as expected. If it's only my IDE causing issues, then no need to build a workaround.

nmohr192:
@nicornk @jonas-w  all comments addressed, ready for final review.
Is the changelog commit going to be part of this PR?

I will commit this here then
```

## [2.1.21] - 2025-10-28

## Added
- add official foundry-platform-sdk support under FoundryContext for public v2 APIs (#112)
```

as: `feat: release v2.1.21`

libs/foundry-dev-tools/src/foundry_dev_tools/public_sdk/auth.py
nmohr192: @nicornk I re-implemented it as you suggested initially for now. It's just a tiny class anyways with no dependencies...

jonas-w: @nmohr192 alright I think I found at least an improvement.

You still can't import like this: `from foundry_dev_tools._optional.foundry_platform_sdk.foundry_sdk._core.auth_utils import Token`

But at least you can use IDE features/autocompletion.

```
from __future__ import annotations

try:
    import foundry_sdk
    import foundry_sdk.v2

    foundry_sdk.__fake__ = False
except ImportError:
    from foundry_dev_tools._optional import FakeModule

    foundry_sdk = FakeModule("foundry_sdk")

__all__ = ["foundry_sdk"]
```

If you just add all the extra imports to the _optional file, in this case for example `import foundry_sdk.v2`

Can you confirm that this would solve IDE autocompletion in your setup too?

I need to probably do this also in other parts of the project, so it works in vscode.

nmohr192: @jonas-w ah, amazing find! Works :)

jonas-w: @nmohr192 awesome, sorry about the whole _optional thing!

libs/foundry-dev-tools/src/foundry_dev_tools/config/context.py
```diff
@@ -184,6 +186,29 @@ def foundry_rest_client(self) -> FoundryRestClient:
...
 
+    @cached_property
+    def public_client_v2(self) -> FoundrySdkV2Client:
+        """Returns :py:class:`foundry_sdk.v2.FoundryClient`.
+
+        Uses fdt context to integrate with the official foundry-platform-sdk.
+
+        Examples:
+            >>> from foundry_dev_tools._optional.polars import pl
+            >>> from foundry_dev_tools._optional.pyarrow import pa
```
jonas-w: I think using the direct imports in examples is okay IMO.
`import polars as pl`...

nmohr192: yeah, good point. Copy & pasted from some test code..

jonas-w (approved):

PR #115, opened 2026-01-17
Implement query_foundry_sql(return_type='polars')

bernhardschaefer:
# Summary

Extend `query_foundry_sql()` in the classes `Dataset`, `FoundryRestClient`, and `FoundrySqlServerClient`  with the return_type polars.

# Checklist

### DIFF
--- a/libs/foundry-dev-tools/src/foundry_dev_tools/clients/data_proxy.py
+++ b/libs/foundry-dev-tools/src/foundry_dev_tools/clients/data_proxy.py
@@ -27,6 +27,7 @@ if TYPE_CHECKING:
     from pathlib import Path
 
     import pandas as pd
+    import polars as pl
     import pyarrow as pa
     import pyspark
     import requests
@@ -111,7 +112,7 @@ class DataProxyClient(APIClient):
         branch: Ref = ...,
         sql_dialect: SqlDialect = ...,
         timeout: int = ...,
-    ) -> pd.core.frame.DataFrame: ...
+    ) -> pd.DataFrame: ...
 
     @overload
     def query_foundry_sql_legacy(
@@ -137,6 +138,16 @@ class DataProxyClient(APIClient):
     def query_foundry_sql_legacy(
         self,
         query: str,
+        return_type: Literal["polars"],
+        branch: Ref = ...,
+        sql_dialect: SqlDialect = ...,
+        timeout: int = ...,
+    ) -> pl.DataFrame: ...
+
+    @overload
+    def query_foundry_sql_legacy(
+        self,
+        query: str,
         return_type: Literal["raw"],
         branch: Ref = ...,
         sql_dialect: SqlDialect = ...,
@@ -151,7 +162,7 @@ class DataProxyClient(APIClient):
         branch: Ref = ...,
         sql_dialect: SqlDialect = ...,
         timeout: int = ...,
-    ) -> tuple[dict, list[list]] | pd.core.frame.DataFrame | pa.Table | pyspark.sql.DataFrame: ...
+    ) -> tuple[dict, list[list]] | pd.DataFrame | pl.DataFrame | pa.Table | pyspark.sql.DataFrame: ...
 
     def query_foundry_sql_legacy(
         self,
@@ -160,7 +171,7 @@ class DataProxyClient(APIClient):
         branch: Ref = "master",
         sql_dialect: SqlDialect = "SPARK",
         timeout: int = 600,
-    ) -> tuple[dict, list[list]] | pd.core.frame.DataFrame | pa.Table | pyspark.sql.DataFrame:
+    ) -> tuple[dict, list[list]] | pd.DataFrame | pl.DataFrame | pa.Table | pyspark.sql.DataFrame:
         """Queries the dataproxy query API with spark SQL.
 
         Example:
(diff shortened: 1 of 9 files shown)


libs/foundry-dev-tools/src/foundry_dev_tools/clients/foundry_sql_server.py
```diff
@@ -105,75 +116,81 @@ def query_foundry_sql(
...
-                    return_type=return_type,
+                if getattr(pl, "__fake__", False):
+                    msg = "The optional 'polars' package is not installed. Please install it to request polars results."
+                    raise ImportError(msg)  # noqa: TRY301
+
+                arrow_table = arrow_stream_reader.read_all()
+                return pl.from_arrow(arrow_table)
+            if return_type == "arrow":
+                return arrow_stream_reader.read_all()
+        except (
```
nicornk: I went through this function and I believe the try catch wrapping can be removed.

libs/foundry-dev-tools/src/foundry_dev_tools/clients/foundry_sql_server.py
```diff
@@ -105,75 +116,81 @@ def query_foundry_sql(
...
-            try:
-                response_json = self.api_queries_execute(
-                    query,
-                    branch=branch,
-                    dialect=sql_dialect,
-                    timeout=timeout,
-                ).json()
+        assert_in_literal(return_type, SQLReturnType, "return_type")
+
+        if return_type == "raw":
```
nicornk: Moving this up will break the fallback that is tested in test_legacy_fallback defined in `foundry-dev-tools/tests/integration/clients/test_foundry_sql_server.py`

I introduced this so data can be queried in constrained environments where `arrow` is not installed.

bernhardschaefer: Good catch. I also realized this is related to the try...catch as well. In the original code, if return_type != "raw" and it runs into either FoundrySqlSerializationFormatNotImplementedError or ImportError and the return_type is also != "arrow", then it will just swallow the exception and fall back to query_foundry_sql_legacy.

My suggestion: I will only add the if return_type == polars part and keep everything else as it was before but add some comments to document this behavior. What do you think?

nicornk: image
agree - make sure this is deleted. it's actually dead-code that will not even work.

libs/foundry-dev-tools/src/foundry_dev_tools/clients/foundry_sql_server.py
```diff
@@ -75,16 +86,16 @@ def query_foundry_sql(
         branch: Ref = ...,
         sql_dialect: SqlDialect = ...,
         timeout: int = ...,
-    ) -> tuple[dict, list[list]] | pd.core.frame.DataFrame | pa.Table | pyspark.sql.DataFrame: ...
+    ) -> tuple[dict, list[list]] | pd.DataFrame | pl.DataFrame | pa.Table | pyspark.sql.DataFrame: ...
 
-    def query_foundry_sql(
+    def query_foundry_sql(  # noqa: C901
```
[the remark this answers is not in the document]
bernhardschaefer: C901 should have already been triggered by pre-commit in HEAD version and is unrelated to this change.

libs/foundry-dev-tools/src/foundry_dev_tools/clients/foundry_sql_server.py
```diff
@@ -105,75 +116,81 @@ def query_foundry_sql(
...
+            if return_type == "polars":
+                from foundry_dev_tools._optional.polars import pl
 
-                return self._query_fsql(
-                    query=query,
-                    branch=branch,
-                    return_type=return_type,
+                if getattr(pl, "__fake__", False):
+                    msg = "The optional 'polars' package is not installed. Please install it to request polars results."
+                    raise ImportError(msg)  # noqa: TRY301
```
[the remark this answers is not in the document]
bernhardschaefer: good rid of explicit raise ImportError as this is already thrown by FakeModule

libs/foundry-dev-tools/src/foundry_dev_tools/clients/foundry_sql_server.py
bernhardschaefer: Changes:
- got rid of the dead code
- documented the fallback logic to legacy client (I now understand that ImportErrors are indirectly raised in FakeModule when trying to access an attribute of the fake module)
- implemented polars in queray_foundry_sql_legacy so that the final uncaught ImportError will be raised at this point

libs/foundry-dev-tools/src/foundry_dev_tools/clients/foundry_sql_server.py
bernhardschaefer: try catch logic is required so that fallback to legacy client still works, as discussed in other thread.

libs/foundry-dev-tools/src/foundry_dev_tools/clients/data_proxy.py
```diff
@@ -206,20 +217,29 @@ def query_foundry_sql_legacy(
         response_json = response.json()
         if return_type == "raw":
             return response_json["foundrySchema"], response_json["rows"]
+        # return_type arrows, pandas and polars use the FakeModule implementation in
```
nicornk:
```suggestion
        # return_type arrow, pandas and polars use the FakeModule implementation in
```

PR #118, opened 2026-03-02
Feat/oauth cli

nicornk:
# Summary

# Checklist

- [ ] You agree with our [CLA](https://gist.githubusercontent.com/emdgroup-admin/16cc45ea4315c2ef29eb9d9afc36fcf5/raw/abb9c91f15278a62b9ac3e66144bcd27fa9485c2/CLA.md)
- [ ] Used [pre-commit hooks](https://emdgroup.github.io/foundry-dev-tools/develop.html#pre-commit-hooks-formatting) to format and lint the code.


libs/oauth-cli/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+version = "0.1.0"
+edition = "2021"
+description = "Standalone OAuth2 CLI for Palantir Foundry"
+
+[[bin]]
+name = "foundry-dev-tools-oauth"
+path = "src/main.rs"
+
+[dependencies]
+clap = { version = "4", features = ["derive"] }
```
dstoeckel, line 12: Clap is notorious for being large and slow to build. If you are not using advanced features try something simpler such as https://docs.rs/argh/0.1.14/argh/

libs/oauth-cli/src/server.rs
```diff
@@ -0,0 +1,250 @@
...
+<head><title>Authorization Failed</title></head>
+<body style="font-family: sans-serif; text-align: center; padding: 50px;">
+  <h1>Authorization failed</h1>
+  <p>Missing authorization code. Please try again.</p>
+</body>
+</html>"#;
+
+/// A bound callback server ready to accept the OAuth redirect.
+pub struct CallbackServer {
+    server: tiny_http::Server,
```
dstoeckel, line 32: You are using tiny_http here, which is likely fine functionality wise. However, you already are depending on reqwest, which is bringing in hyper that also allows you to write a server.

libs/oauth-cli/src/cache.rs
dstoeckel, line 1: The approach in this file works, but for idiomatic rust code, you would:
* Make the cache / keyring access a trait
* Implement that trait for the linux platform and for the mac/windows platform
* Switch to the appropriate implementation using a cfg! switch

libs/oauth-cli/src/log.rs
dstoeckel, line 1: Consider using the log crate for this. It is a generic logger facade, the de-facto standard, and can have multiple backends (including very simple ones).

libs/oauth-cli/src/cache.rs
```diff
@@ -0,0 +1,525 @@
...
+}
+
+/// Execute a closure while holding an exclusive file lock.
+pub fn with_lock<F, T>(config_dir: &Path, debug: bool, f: F) -> Result<T>
+where
+    F: FnOnce() -> Result<T>,
+{
+    ensure_config_dir(config_dir)?;
+
+    let lock_path = lock_file_path(config_dir);
```
dstoeckel, line 294: This cache is all or nothing. If I am understanding the implementation correctly, you'll never write more than a single file, so it would be possible to have a more granular locking mechanism in place that creates a dedicated .lock file for each token.

libs/oauth-cli/src/cli.rs
```diff
@@ -0,0 +1,349 @@
...
+    let has_token = cache::load(
+        &config.config_dir,
+        &config.hostname,
+        &config.client_id,
+        &config.scopes,
+        config.debug,
+    )?
+    .is_some();
+
+    eprintln!("  Hostname:  {}", config.hostname);
```
dstoeckel, line 306: Why print this to stderr? If it is supposed to work for machines, you would probably want some JSON output anyway.

libs/oauth-cli/src/config.rs
```diff
@@ -0,0 +1,387 @@
...
+    if hostname.is_empty() {
+        return Err(Error::InvalidHostname {
+            hostname: hostname.to_string(),
+        });
+    }
+
+    // Must contain only valid domain characters: alphanumeric, hyphens, dots
+    let is_valid = hostname
+        .chars()
+        .all(|c| c.is_ascii_alphanumeric() || c == '-' || c == '.');
```
dstoeckel, line 75: Note that ascii may be overly restrictive. Domain names may very well include unicode: https://www.rfc-editor.org/rfc/rfc5890

libs/oauth-cli/src/config.rs
```diff
@@ -0,0 +1,387 @@
...
+    candidates[0].join("oauth")
+}
+
+/// Validate that the hostname is a plain domain name, not a URL or path-injected string.
+/// Rejects any hostname containing characters that could cause URL injection when
+/// interpolated into `https://{hostname}/multipass/api/...`.
+fn validate_hostname(hostname: &str) -> Result<()> {
+    if hostname.is_empty() {
+        return Err(Error::InvalidHostname {
+            hostname: hostname.to_string(),
```
dstoeckel, line 68: That's a bit weird. Hostname is empty after all ...

libs/oauth-cli/src/oauth.rs
```diff
@@ -0,0 +1,228 @@
...
+    pub access_token: String,
+    pub refresh_token: Option<String>,
+}
+
+/// Characters allowed in a PKCE code_verifier (RFC 7636 §4.1).
+const PKCE_CHARS: &[u8] = b"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-._~";
+
+/// Generate a PKCE code_verifier and code_challenge (S256).
+pub fn generate_pkce() -> Pkce {
+    let mut rng = rand::rng();
```
dstoeckel, line 28: This should have a comment that rand::rng returns a CSPRNG and is, in fact, safe to use here.

libs/oauth-cli/src/oauth.rs
```diff
@@ -0,0 +1,228 @@
...
+    let verifier: String = (0..128)
+        .map(|_| {
+            let idx = rng.random_range(0..PKCE_CHARS.len());
+            PKCE_CHARS[idx] as char
```
dstoeckel, line 32: Why cast to char here, when the next thing that is being done is to cast to bytes again?

libs/oauth-cli/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+reqwest = { version = "0.13", default-features = false, features = ["json", "blocking", "native-tls", "form"] }
+serde = { version = "1", features = ["derive"] }
+serde_json = "1"
+sha2 = "0.10"
+base64 = "0.22"
+rand = "0.10"
+tiny_http = "0.12"
+open = "5"
+dirs = "6"
+url = "2"
```
dstoeckel, line 22: You are including url here, but you implement host parsing and validation on your own. Why not use https://docs.rs/url/2.5.8/url/enum.Host.html#impl-Host to perform the parsing and then checking the enum whether it is a valid domain name (and not an IP address)

libs/oauth-cli/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+chrono = { version = "0.4", features = ["serde"] }
+fd-lock = "4"
```
dstoeckel, line 24: This dependency seems unused. Either use it and discard the homebrew locking code or remove it.

libs/oauth-cli/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+keyring = { version = "3", features = ["apple-native", "windows-native"] }
```
dstoeckel, line 25: This should be a platform dependent feature and pulled in during compile on the required platforms only (cfg check). See comment in the cache.rs implementation.

libs/oauth-cli/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "foundry-dev-tools-oauth-cli"
+version = "0.1.0"
+edition = "2021"
```
dstoeckel, line 4: Why not the 2024 edition? This is out for more than a year now.

dstoeckel (commented):
These is my first set of comments. Will do another pass later / when updated.

libs/oauth-cli/src/cache.rs
nicornk: All cached tokens share a single cache file and I don't want race conditions between them.

libs/oauth-cli/src/server.rs
nicornk: We evaluated replacing tiny_http with hyper (which reqwest pulls in transitively). However, reqwest only depends on hyper as a client — hyper's server functionality requires adding explicit dependencies (hyper with the server and http1 features, plus hyper-util with tokio-server). This would also pull in a tokio runtime for what is a single synchronous blocking-wait-for-one-request. tiny_http gives us a simple synchronous recv_timeout API that fits this use case well without introducing async complexity. We're keeping it for now.

libs/oauth-cli/Cargo.toml
nicornk: Just as slow as this Github page :D replaced it with argh

